PR #46, opened 2022-10-27
Highlight anomalies in long run statistics

quirogas:
- Add highlighting of anomalies for long collection data.
- Fix Module imports 
- Fix tests resource path

Resolves:
- [highlight object growth](https://jira.eng.vmware.com/browse/VOPERATION-35572)
- [highlight property and metric growth](https://jira.eng.vmware.com/browse/VOPERATION-35576)
- [highlight Property value growth](https://jira.eng.vmware.com/browse/VOPERATION-35574)
- [highlight event growth](https://jira.eng.vmware.com/browse/VOPERATION-35577)

TODO:
-  [highlight relationship growth](https://jira.eng.vmware.com/browse/VOPERATION-35575)


tests/validation/test_highlights.py
```diff
@@ -0,0 +1,366 @@
...
+            response.text = json.dumps(_json)
+            collection_bundle = CollectionBundle(pytest.REQUEST, copy.deepcopy(response), pytest.DURATION,
+                                                 pytest.CONTAINER_STATISTICS)
+            collection_bundle.collection_number = collection
+            growing_unique_objects_overtime.add(collection_bundle)
+            _json["result"][0]["key"]["name"] = f"{collection}"
+
+        highlight = highlight_object_growth(growing_unique_objects_overtime.long_collection_statistics)
+        assert (ResultLevel.WARNING,
+                "Object of type HighlightsMP::Growing Object displayed growth of 9.65%") in highlight.messages
```
kjrokos, line 103: A few things here that apply to all the messages:
* "Object of type" should be "Objects of type", since we are talking about all objects of the given type, not a singular one
* Should add single quotes around types, keys, etc
* I think we talked about adding a time period to the unit, so it's clear what the growth rate is measuring.
```
"Objects of type 'HighlightsMP::Growing Object' had a growth rate of 9.65% per ???"
```

tests/validation/test_highlights.py
```diff
@@ -0,0 +1,366 @@
...
+            response.text = json.dumps(_json)
+            collection_bundle = CollectionBundle(pytest.REQUEST, copy.deepcopy(response), pytest.DURATION,
+                                                 pytest.CONTAINER_STATISTICS)
+            collection_bundle.collection_number = collection
+            growing_unique_objects_overtime.add(collection_bundle)
+            if collection % 3 == 0:
+                _json["result"][0]["key"]["name"] = f"{collection}"
+
+        highlight = highlight_object_growth(growing_unique_objects_overtime.long_collection_statistics)
+        assert (ResultLevel.WARNING,
```
kjrokos, line 120: Do warnings get displayed by default? If not I wonder if we should change that.

kjrokos: I'm not sure ERROR is appropriate here, but I don't think they should be hiding in a log by default.

tests/validation/test_highlights.py
```diff
@@ -0,0 +1,366 @@
...
+                                                 pytest.CONTAINER_STATISTICS)
+            collection_bundle.collection_number = collection
+            increasing_number_of_objects_per_collection.add(collection_bundle)
+            new_object = copy.deepcopy(pytest.JSON["result"][0])
+            new_object["key"]["name"] = f"{collection}"
+            _json["result"].append(new_object)
+
+        highlight = highlight_object_growth(increasing_number_of_objects_per_collection.long_collection_statistics)
+        assert (ResultLevel.WARNING,
+                "Object of type HighlightsMP::Growing Object displayed growth of 9.65%") in highlight.messages
```
kjrokos, line 157: This is the other scenario compared to the previous cases, correct? If so we should make sure the messages in each case are different and accurately describe the scenario we are detecting.

tests/validation/test_highlights.py
```diff
@@ -0,0 +1,366 @@
...
+                "objectKind": "Test Child"
+            }]
+
+            _json["result"] = [parent_object, child_object]
+            _json["relationships"] = [{"parent": parent_relationship, "children": child_relationship}]
+
+        highlight = highlight_relationship_growth(
+            increasing_number_of_property_values_per_collection.long_collection_statistics)
+        assert (ResultLevel.WARNING,
+                "Object of type HighlightsMP::Test Parent displayed relationship growth of 2.81%") in highlight.messages
```
kjrokos, line 312: We may want to confirm with the platform team how relationships are processed from the API. In the original SDK, it was a problem if relationships changed too much, even if they weren't growing. For example, if a resource is constantly switching between several parents, the number of unique relationships won't change much but if there were enough relationship changes that could be a problem.

vrealize_operations_integration_sdk/stats.py
```diff
@@ -54,16 +54,21 @@ def get_average(inputs: list):
     return sum(inputs) / len(inputs)
 
 
-def get_growth_rate(inputs: list):
-    present = inputs[-1]
-    past = inputs[0]
-
+# TODO: change signature to initial, final, and time
```
kjrokos, line 57: TODO is done

vrealize_operations_integration_sdk/stats.py
```diff
@@ -54,16 +54,21 @@ def get_average(inputs: list):
...
+def get_growth_rate(initial, final, time):
+    """
+    Calculates average percent growth rate per hour
+    :param initial: the initial number of objects
+    :param final: the final number of objects / things
+    :param time: the total time period in seconds
+    :return: float that represents the percent of g
```
kjrokos, line 64: time is specified as seconds, but description says growth rate per hour -- I don't see any conversion between them, though.

I'd recommend changing the `time` parameter to include the unit, e.g., `get_growth_rate(initial, final, duration_in_seconds)` if seconds is correct.

vrealize_operations_integration_sdk/ui.py
```diff
@@ -379,3 +379,10 @@ def __repr__(self):
             output += formatting.format(*row) + "\n"
 
         return output
+
+
+# TODO: write highlight class
+class Highlight:
+    def __init__(self, bundle, condition):
+
+        pass
```
kjrokos, line 388: Should either implement this or remove it

vrealize_operations_integration_sdk/validation/highlights.py
```diff
@@ -0,0 +1,98 @@
...
+    return highlights
+
+
+def highlight_property_value_growth(long_collection_statistics: LongCollectionStatistics) -> Result:
+    objects_with_property_value_growth = [(object_type, stats.property_values_growth_rate) for
+                                          object_type, stats in
+                                          long_collection_statistics.long_object_type_statistics.items()
+                                          if stats.property_values_growth_rate > 0]
+
+    MAX_NUMBER_OF_PROPERTY_VALUES = 256  # This is mostly subjective.
```
kjrokos, line 54: That is probably too high. This actually might be a case where it's useful to use the number of collections to determine the threshold, something like `min(32, number_of_collections)` -- that is, if we see a new value every collection that's probably a red flag even if there are only a few collections. We've very rarely (if ever) used an enum with more than 32 values, and even when we have they shouldn't change often enough to trigger anything. Usually a few values are very common and the rest are exceptional cases.

kjrokos: Thinking about this some more, there may also be some dependency on the number of objects the property exists on as well. For example, we might have an OS property that is slightly different for (because it has a build number, for example, or maybe some hardware identifier), but never changes. That should be OK as well.

tests/validation/test_highlights.py
kjrokos: My read of their response is that this is something we will want to detect.

vrealize_operations_integration_sdk/validation/highlights.py
```diff
@@ -0,0 +1,97 @@
...
+    threshold = get_growth_rate(1, MAX_NUMBER_OF_PROPERTY_VALUES, long_collection_statistics.long_run_duration)
+
+    highlights = Result()
+    if len(objects_with_property_value_growth):
+        for obj_type, growth in objects_with_property_value_growth:
+            if growth > threshold:
+                highlights.with_error(
+                    f"Objects of type '{obj_type}' displayed excessive property value growth of {growth:.2f}% per hour.")
+            else:
+                highlights.with_warning(f"Objects of type '{obj_type}' displayed property value growth of {growth:.2f}% per hour.")
```
kjrokos, line 64: One thing that we're missing with all of the warnings is some explanation. We show what the problem is, but not why it's a problem or what to look for to fix the issue. I think for anyone who is not as familiar with some of these issues as we are this will be very important context to provide.

tests/validation/test_highlights.py
kjrokos: For posterity: After some discussion, we realized that the case where the number of returned objects grows continuously typically happened in the past due to data structures not being cleared properly between collections. For the SDK, the adapter process does not persist between collections, so this type of error should be much less common. Thus, we are not going to specifically call out this case. If it does happen for some reason, it will still be caught as it is a special case of the general case where the number of unique objects over time is growing.

vrealize_operations_integration_sdk/collection_statistics.py
```diff
@@ -78,6 +81,55 @@ def get_summary(self):
...
+        return get_growth_rate(self.events_stats.data_points[0],
+                               self.events_stats.data_points[-1],
+                               hours)
+
+    @LazyAttribute
+    def relationships_growth_rate(self):
+        hours = self.long_run_duration / 3600
+        return get_growth_rate(self.relationships_stats.data_points[0],
+                               self.relationships_stats.data_points[-1],
+                               hours)
```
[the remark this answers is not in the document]
kjrokos: The period (i.e., the time between collections) shouldn't make much of a difference if we did it correctly. I think you meant the duration of the long run. If that is short and we have a small amount of growth, that could certainly lead to large numbers for the growth rate.

One thing that we could potentially add is the absolute number - for example, "The number of metrics on objects of type 'Object' grew by 10 over the duration of the long run (1100% per hour)."

We can discuss later. If we want to do this it should be a subsequent PR.

vrealize_operations_integration_sdk/validation/highlights.py
```diff
@@ -0,0 +1,97 @@
...
+def highlight_object_growth(long_collection_statistics: LongCollectionStatistics) -> Result:
+    objects_with_growth = [(object_type, stats.objects_growth_rate) for
+                           object_type, stats in
+                           long_collection_statistics.long_object_type_statistics.items()
+                           if stats.objects_growth_rate > 0]
+
+    highlights = Result()
+    if len(objects_with_growth):
+        for obj_type, growth in objects_with_growth:
+            highlights.with_warning(f"Objects of type '{obj_type}' displayed growth of {growth:.2f}% per hour.")
```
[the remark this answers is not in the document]
kjrokos: Looks like you changed all of them except the one here (object growth)

vrealize_operations_integration_sdk/validation/highlights.py
kjrokos: I like how these are reading. However, there are a few things that are incorrect in the explanations. Let's go through them one by one when you get some time.

kjrokos (approved):

PR #50, opened 2022-11-04
Add xml fragments to programmatically-generated describe.xml

kjrokos:
This will allow us to put xml files in the 'conf' directory that, if they have the correct format, get added into the programmatically generated describe from the adapter.

I intentionally did not add this to the case where the describe.xml file is _not_ programmatically generated, but we could do so in the future with some (relatively small amount of) additional work. However, there are some limitations, namely that multiple languages are not supported - which may be one of the few remaining use cases for having a completely manually-generated describe.xml file now.

### DIFF
--- a/vrealize_operations_integration_sdk/describe.py
+++ b/vrealize_operations_integration_sdk/describe.py
@@ -5,10 +5,9 @@ from __future__ import annotations
 import json
 import logging
 import os
-from collections import OrderedDict
-
-from xml.etree.ElementTree import Element, SubElement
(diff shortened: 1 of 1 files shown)


MarkArbogast:
I don't see much marginal value in supporting this if the MP developer is already hand-writing their describe.xml.

MarkArbogast:
Looks good. It will be good to explain this feature in an advanced section of the README and maybe provide some snippets / examples that developers can reference.

MarkArbogast (approved):

PR #53, opened 2022-11-17
Property value functions

quirogas:
- Add value functions to Property Object:
   - get_last_property_value
   - get_property_values
   
- Add value functions to Metric Object:
   - get_last_metric_value
   - get_metric_values
    
- Fix Property and Metric timestamps

### DIFF
--- a/lib/python/src/aria/ops/object.py
+++ b/lib/python/src/aria/ops/object.py
@@ -231,6 +231,33 @@ class Object:
         """
         return list(filter(lambda metric: metric.key == key, self._metrics))
 
+    def get_metric_values(self, key) -> list[float]:
+        """
+
+        :param key: Metric key of the metric to return.
+        :return: A list of the metric values in chronological order.
+        """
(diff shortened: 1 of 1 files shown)


lib/python/src/aria/ops/object.py
```diff
@@ -264,6 +264,34 @@ def get_property(self, key) -> list[Property]:
...
+        properties.sort(key=lambda property: property.timestamp, reverse=True)
+
+        return [p.value for p in properties]
+
+    def get_last_property_value(self, key) -> str | None:
+        """
+
+        :param key: Property key of the property to return.
+        :return:  return the last value of the property with the matching key or None
+        if the property Key doesn't match any property
```
kjrokos, line 286:
```suggestion
        :param key: Property key of the property value to return.
        :return:  The latest value of the property or None if no property exists with the given key.
```

kjrokos (approved):

PR #63, opened 2023-01-12
Feature/automated pak file build

quirogas:
- Add parameters to allow mp-build to run in a CI/CD environment 
-  Remove the spinner when not running on a terminal
### Resolves:
- [VOPERATION-33248](https://jira.eng.vmware.com/browse/VOPERATION-33248)

### Before (It didn't work):
image

### After:
image


vmware_aria_operations_integration_sdk/mp_build.py
```diff
@@ -136,20 +136,28 @@ def registry_prompt(default: str) -> str:
...
+    if docker_registry is None and docker_registry_arg is None:
         print(
             "mp-build needs to configure a Docker registry to store the adapter container image.",
             "class:information",
         )
-        docker_registry = registry_prompt(
-            default=f"harbor-repo.vmware.com/vmware_aria_operations_integration_sdk_mps/{adapter_kind_key.lower()}"
-        )
+        docker_registry = registry_prompt(default=default_registry_value)
+    elif docker_registry_arg is not None and not len(docker_registry_arg):
```
kjrokos, line 155: If this is provided but empty, _and_ there is an entry in the config file, should we still use the default?

I wonder if it might be better to have a special argument string like "default" and not support passing `--registry-tag` with no arguments

vmware_aria_operations_integration_sdk/mp_build.py
```diff
@@ -380,6 +394,18 @@ def main() -> None:
...
+            "--registry-tag",
+            help="The full Docker registry tag where the container image will be stored",
+            nargs="?",
+            const="",
+        )
+
+        parser.add_argument("--registry-username", help="The Docker registry username")
+
+        parser.add_argument("--registry-password", help="The Docker registry password")
+
```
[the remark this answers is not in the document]
kjrokos: Should also update the `mp-build.md` documentation with the new arguments. Would probably be good to add a section on what is required to build from a headless node.

pyproject.toml
```diff
@@ -44,7 +44,7 @@ jsonschema-spec = "^0.1.2"
 validators = "^0.20.0"
 
 [tool.poetry.group.dev.dependencies]
-setuptools = "^65.6.3"
+# setuptools = "^65.6.3"
```
kjrokos, line 47: let's just delete this line

vmware_aria_operations_integration_sdk/docker_wrapper.py
```diff
@@ -25,13 +25,22 @@
 from vmware_aria_operations_integration_sdk.constant import DEFAULT_PORT
 from vmware_aria_operations_integration_sdk.stats import convert_bytes
 from vmware_aria_operations_integration_sdk.stats import LongRunStats
-from vmware_aria_operations_integration_sdk.threading import threaded
+from vmware_aria_operations_integration_sdk.temp_threading import threaded
```
kjrokos, line 28: This should be changed back. If it's still giving you issues let's open a bug to rename it in a non-temporary way

kjrokos (approved):

PR #66, opened 2023-01-18
Add additional logic for mp-build parameters

quirogas:
- Fix a bug where mp-build would prioritize parameter value over config file, even when the parameter was None.
- Fix an error where docker mp-build would crash when the user didn't pass docker username or password.


vmware_aria_operations_integration_sdk/mp_build.py
```diff
@@ -146,6 +146,9 @@ def get_docker_registry(
     default_registry_value = f"harbor-repo.vmware.com/vmware_aria_operations_integration_sdk_mps/{adapter_kind_key.lower()}"
 
     original_value = docker_registry
+    print(
```
kjrokos, line 149: Is this intended to be here, or was it just for debugging?

If intended, there is a typo in `regitry` -> `registry`, and should remove 'docker' in favor of 'container'.

kjrokos (approved):

PR #71, opened 2023-02-02
Story/library relationship improvements

kjrokos:
Adds a number of features to the Python library:
* If an AdapterDefinition is passed into a 'CollectResult', external object types
  are no longer returned in the result unless they contain content. Relationships
  are not affected. (Note: requires unreleased mp-test update)
* Adds 'RelationshipUpdateModes' enum to control how relationships are returned from 
  'CollectResult'
* Add 'get_object' and 'get_objects_by_type' helper methods in 'CollectResult'
* Add 'adapter_type()' and 'object_type()' helper methods in 'Object'
* Add 'has_content()' method in 'Object' that returns true if the object contains 
  at least one metric, property, or event

In addition
* Adds unit tests for CollectResult class
* Fixes a bug in mp-test that occurs when collect returns an object in the "relationships" dictionary that is not part of the "result" dictionary

Resolves: https://jira.eng.vmware.com/browse/VOPERATION-37456, https://jira.eng.vmware.com/browse/VOPERATION-37232

Note: Failing CI will be resolved after #70 is merged.

### DIFF
--- a/lib/python/src/aria/ops/result.py
+++ b/lib/python/src/aria/ops/result.py
@@ -128,10 +128,10 @@ class RelationshipUpdateModes(Enum):  # type: ignore
     there are relationships between objects in the Result. All currently-existing
     relationships in VMware Aria Operations will be preserved.
 
-    If 'update_relationships' is 'AUTO' (or not explicitly set), then if any object
-    in the result has at least relationship, the mode will behave like 'ALL',
-    and if no objects have any relationships in the Result, the mode will behave like
-    'NONE'. This default behavior makes it easy to skip collecting all relationships
+    If 'update_relationships' is 'AUTO' (or not explicitly set), then the mode will
+    behave like 'ALL' if any object in the Result has at least one relationship,
+    otherwise the mode will behave like 'NONE' if no objects have any relationships in
+    the Result. This default behavior makes it easy to skip collecting all relationships
     for a collection without overwriting previously-collected relationships, e.g., for
     performance reasons.
 
@@ -141,7 +141,7 @@ class RelationshipUpdateModes(Enum):  # type: ignore
     'add_children'), existing child relationships in VMware Aria Operations will be
     preserved. This means that to remove all relationships from an object
     (without setting any new relationships), the adapter must call 'add_children' on
-    the object with an empty collection of children. This is useful for updating a
+    the object with an empty collection of children. This mode is useful for updating a
     subset of objects' relationships in a collection, but requires more care to
     ensure relationships are removed when appropriate.
     """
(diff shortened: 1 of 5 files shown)


lib/python/src/aria/ops/result.py
```diff
@@ -111,6 +115,44 @@ def send_results(self, output_pipe: str = sys.argv[-1]) -> None:
...
+    relationships_.
+
+    If 'update_relationships' is 'PER_OBJECT', then only objects with updated
+    relationships will be returned. Note: Any object that has not been updated
+    will retain existing relationships. This means that to remove all
+    relationships from an object (without setting any new relationships), the
+    adapter must call 'add_children' on the object with an empty collection of
+    children.
+
+    If an object's relationships are returned, but the object has no current
```
MarkArbogast, line 139: I'm having trouble parsing this sentence.

MarkArbogast:
Just one question, otherwise looks good! Great to see the new unit tests.

lib/python/src/aria/ops/result.py
kjrokos: Hmm, I see - I think we need some better terminology. It turns out that this:

```
relationships: []
```
and this:
```
relationships: [
   {
       parent: { ... }
       children: []
   }
]
```
are not equivalent. The first one returns no relationship objects, and the second returns a relationship object that has no relationships. The first one does nothing, and the second one removes any existing children relationships to the specified parent (if there were any). The sentence is referring to the second case.

Do you have any wording suggestions?

lib/python/src/aria/ops/result.py
```diff
@@ -111,6 +115,43 @@ def send_results(self, output_pipe: str = sys.argv[-1]) -> None:
...
+    If 'update_relationships' is 'ALL', all relationships between objects are
+    returned. This mode will remove any currently-existing relationships in VMware
+    Aria Operations that are not present in the Result.
+
+    If 'update_relationships' is 'NONE', no relationships will be returned, even if
+    there are relationships between objects in the Result. All currently-existing
+    relationships in VMware Aria Operations will be preserved.
+
+    If 'update_relationships' is 'AUTO' (or not explicitly set), then if any object
+    in the result has at least relationship, the mode will behave like 'ALL',
```
MarkArbogast, line 132: at least relationship -> at least **one** relationship

lib/python/src/aria/ops/result.py
MarkArbogast: much better now!

lib/python/src/aria/ops/result.py
kjrokos: Updated

PR #77, opened 2023-02-24
Adds an optional 'default_value' parameter to get_identifier()

kjrokos:
I was surprised this was missing.

Please think about if the behavior in each case makes sense (identifier not present vs identifier has Falsy value).

### DIFF
--- a/lib/python/src/aria/ops/object.py
+++ b/lib/python/src/aria/ops/object.py
@@ -82,15 +82,23 @@ class Key:
     def __hash__(self) -> int:
         return hash(self.__key())
 
-    def get_identifier(self, key: str) -> Optional[str]:
+    def get_identifier(
+        self, key: str, default_value: Optional[str] = None
+    ) -> Optional[str]:
         """Return the value for the given identifier key
 
         :param key: The identifier key
-        :return: The value associated with the identifier, or 'None' if the identifier key does not exist
+        :param default_value: An optional default value
+        :return: The value associated with the identifier.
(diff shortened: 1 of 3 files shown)


lib/python/src/aria/ops/object.py
```diff
@@ -82,15 +82,23 @@ def __eq__(self, other: Any) -> bool:
...
+    ) -> Optional[str]:
         """Return the value for the given identifier key
 
         :param key: The identifier key
-        :return: The value associated with the identifier, or 'None' if the identifier key does not exist
+        :param default_value: An optional default value
+        :return: The value associated with the identifier.
+        If the value associated with the identifier is empty and 'default_value' is
+        provided, returns 'default_value'.
+        If the identifier does not exist, returns default_value if provided, else 'None'.
```
MarkArbogast, line 95: how should we behave if an adapter instance identifier definition has a default value, vs. if it does not?

kjrokos: Good question. IIRC, the only time the default in the definition is used is when creating an adapter instance, the default is populated in the UI. If the user doesn't modify it, then it becomes the identifier value.

I can imagine that in most cases the default_value here and the one in the definition will be the same, but at least right now we have no way to automatically do that. Also, there may be edge cases where they should not be the same, e.g., if the default populated in the UI is a URL, but a blank URL is also valid - the user can delete the default, and the developer will want to call `get_identifier` with no default value to ensure they can detect that.

PR #78, opened 2023-02-24
Bug/race condition

quirogas:
In some scenarios, especially those with large collections, the reader threat takes additional time to load the JSON, so we assume that the writer threat closed before being able to write the result and attempt to close the Writer FIFO manually. Since the FIFO was already opened and closed, the program will hang forever. To mitigate the issue, we added a timer to allow the reader to process the payload.


kjrokos:
Looks good. I appreciate the thorough comments and logs!

Can we add a ticket to the SDK Jira board (if there isn't one already) for finding a guaranteed solution, and reference it in the comments?

MarkArbogast (approved):
nice!

kjrokos (approved):

PR #98, opened 2023-03-29
Create and use a custom context for docker builds

### LINKED ISSUE #89: Docker builds take too long in some cases
### Describe the bug

In some cases, running `mp-test` or `mp-build` can take minutes to build the image, while running `docker build .` completes in seconds

### Reproduction steps

This happens when there are very large files in the MP directory, even if they are not part of the docker build.

### Expected behavior

Building the docker image should take approximately the same time in our apps as on the command line.

### Additional context

When we build an image, we tell the build command to build based on the entire MP directory. It turns out, what happens behind the scenes is that the docker process tars the entire directory, and then sends that to the docker REST API's build command ([https://docs.docker.com/engine/api/v1.42/#tag/Image/operation/ImageBuild.](https://docs.docker.com/engine/api/v1.42/#tag/Image/operation/ImageBuild)) This can be a problem if the directory size gets too large, because the process has to tar the entire directory to 'upload' to the docker's rest server, which then has to untar. These operations can be very CPU and memory intensive with large directory sizes.

Fixing #90 will mitigate this issue in many cases.

If we perform the tar operation ourselves, we can control what gets uploaded. We should only need the docker file and app directory - including other directories and files (logs, venv, build, etc) can add a significant amount of time to the build process.

Alternatively, switching from `docker-py` to `Python-on-Whales` should also fix the issue (see #91)

kjrokos:
Creates a custom context to pass to the docker build image command. The custom context excludes any directory that:
  * Is an top-level SDK-created directory that should not be used by the Dockerfile: `build`, ~`conf`~, `content`, `logs`, `resources`
  * Is hidden (e.g., .`git`, `.idea`, etc)
  * Contains a Python virtual environment

Example of running a 2-minute long-run. Times above the 2-minute mark is the time spent setting up the test, the majority of which is building the image.

Without custom context:
```
(venv-test) ❯ du -h -d 0
159M	.
(venv-test) ❯ time mp-test -c Default long-run -d 2m -i 30s
[...]
mp-test -c Default long-run -d 2m -i 30s  14.45s user 1.76s system 11% cpu 2:24.61 total
```

With custom context (Approximately 4MB):
```
(venv-test) ❯ time mp-test -c Default long-run -d 2m -I 30s
[...]
mp-test -c Default long-run -d 2m -i 30s  2.88s user 0.93s system 3% cpu 2:04.67 total
```

Resolves #89

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 VMware Aria Operations Integration SDK
 --------------------------------------
 ## Unreleased
+* Improves container image build times.
 * Do not include '.gitkeep' files in pak files.
 * Adds several flags to mp-build for running on headless build servers:
   * '--no-ttl': Remove UI flourishes and prompts that require a TTL.
--- a/vmware_aria_operations_integration_sdk/docker_wrapper.py
+++ b/vmware_aria_operations_integration_sdk/docker_wrapper.py
@@ -1,9 +1,14 @@
 #  Copyright 2022 VMware, Inc.
 #  SPDX-License-Identifier: Apache-2.0
 import asyncio
+import io
 import json
+import logging
 import os
+import stat
 import subprocess
+import sys
+import tarfile
 import time
 from types import TracebackType
 from typing import Any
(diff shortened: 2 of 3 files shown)


MarkArbogast:
Is it possible a developer would want to read a properties file (or something similar) from the conf directory?

MarkArbogast (approved):
Approved pending consideration of the `conf` directory

kjrokos:
I was going back and forth on what to include/exclude. Traditionally, we used the conf directory to store user-config files (and describe.xml). With a dockerized adapter, though, the conf directory still exists on the CP in the traditional spot, but it isn't mapped to the container. So, we could include the conf directory in the container, but it would be a duplicate of the one on the CP filesystem, and wouldn't have any persistent storage. 

That said, it's likely a small folder size, so it shouldn't affect the build times much, and, especially in the case where the user is _not_ making a programmatic describe, that would be the location the describe.xml lives -- if they want to access it during collection if we don't include the conf directory, then they'd have to duplicate it and maintain two copies.

So, I think you're correct: we should include the conf directory in the image. But definitely an edge case, and I don't know if we should _encourage_ using it.

PR #105, opened 2023-04-05
Windows 10 & 11 Ready

quirogas:
* Fix issue where 'psutil'  to version 5.9.4 to prevent dependency on a template to crash build
* Standardize encoding of properties file to 'utf-8' to prevent encoding errors on Windows

## Before:
image

## After:
image

### DIFF
--- a/vmware_aria_operations_integration_sdk/mp_init.py
+++ b/vmware_aria_operations_integration_sdk/mp_init.py
@@ -427,7 +427,7 @@ def build_project_structure(
         # create template requirements.txt
         requirements_file = os.path.join(path, "adapter_requirements.txt")
         with open(requirements_file, "w") as requirements:
-            requirements.write("psutil==5.9.0\n")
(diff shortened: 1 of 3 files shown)


vmware_aria_operations_integration_sdk/mp_init.py
```diff
@@ -427,7 +427,7 @@ def build_project_structure(
         # create template requirements.txt
         requirements_file = os.path.join(path, "adapter_requirements.txt")
         with open(requirements_file, "w") as requirements:
-            requirements.write("psutil==5.9.0\n")
+            requirements.write("psutil==5.9.4\n")
```
kjrokos, line 430: I had noticed this was an issue recently but didn't look into it at the time. Do you know what caused the older version to stop working? (Was it from when we updated the base OS image?)

kjrokos (approved):

MarkArbogast (approved):

PR #109, opened 2023-04-13
Story/getting stared revision

quirogas:
- Revise and Proofread README.md
    - Add a short section describing Template Adapter
    - Remove project-specific files (python, java, and PowerShell) from the documentation to avoid confusion
    - Add MacOS 13 (Ventura) to the list of supported OS (verified)
- Expand mp-init.md documentation 
    - remove reference to ALB
    - remove the language selection section
    - Add adapter_requirements.txt section
    - Add requirements.txt section
    - Add venv section
    - Re-arrange sections to bring single file information first

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,9 +199,7 @@ the adapter cannot store any data for use in later method calls. Each method is 
     5. Return the CollectResult.
 
 - get_adapter_definition():
-  Defines the object types and attribute types, and in most cases, it also defines the Adapter Instance configuration 
-  parameters/credentials present in a collection. Setting these helps VMware Aria Operations to validate, process, and display the 
-  data correctly.
+  Optional method that defines the Adapter Instance configuration (parameters and credentials used to connect to the target, and configure the management pack) present in a collection, and defines the object types and attribute types present in a collection. Setting these helps VMware Aria Operations to validate, process, and display the data correctly. If this method is omitted, a `describe.xml` file should be manually created inside the `conf` directory with the same data. Generally, this is only necessary when using advanced features of the `describe.xml` file that are not present in this method.
 
 
 For further guidance on using the template project, consult the [Walkthroughs](../README.md#walkthroughs) section.


doc/java_project.md
```diff
@@ -1,23 +0,0 @@
-# Java Project
```
kjrokos, line 1: Let's leave this file. We can replace the contents with something like "Java Project support is planned for a future release". I'd rather not remove entirely because IIRC it's linked from a number of pages.

doc/mp-init.md
```diff
@@ -18,38 +18,38 @@ an initial project structure and create classifiers that other tools and VMware
...
 ```
-Management pack description: This management pack is a proof of concept for the VMware Aria Operations Integration SDK project.
+Management pack description: This management pack is an example for the VMware Aria Operations Integration SDK project.
 ```
 
 #### Vendor
 The name of the vendor/developer of the Management Pack. VMware Aria Operations Manager will display this information during the installation of the  Management Pack generated by this project.
 ```
-Management pack vendor: TVS Team
+Management pack vendor: VMware
```
kjrokos, line 52: I'd suggest something like `MyCompany` here

doc/mp-init.md
```diff
@@ -65,81 +65,71 @@ The _icon_ for the Management Pack generated by this project. The icon will be d
...
-│   ├── recommendations
-│   ├── reports
-│   ├── resources
-│   ├── supermetrics
-│   ├── symptomdefs
-│   └── traversalspecs
 ├── eula.txt
 ├── manifest.txt
+├── requirements.txt
+├── venv
```
kjrokos, line 86: I think the `venv` folder name is now generated using the adapter name

doc/mp-init.md
```diff
@@ -163,14 +153,36 @@ mark. If your editor automatically includes the byte-order mark when saving to U
...
 
 <sup>2</sup> Localized fields are specified in the property files located in the `resources/` directory. See [Adding Localization](adding_content.md#adding-localization).
 
+### venv (directory)
+
+A Python virtual environment for the project. The directory name is prefixed with __venv__ followed by the display name of the management pack. This virtual environment is created to allow users to use different versions of the SDK for every project.
+
+**NOTE:** This directory is excluded from version control. Users cloning the project will have to create their own venv directory([ venv documentaion](https://docs.python.org/3/library/venv.html)).
+
+### app (directory)
```
kjrokos, line 170: This is perhaps the most important directory, and should be at or near the top.

doc/mp-init.md
```diff
@@ -163,14 +153,36 @@ mark. If your editor automatically includes the byte-order mark when saving to U
...
+The conf directory contains configuration files used by VMware Aria Operations to create a management pack instance.
+
+- describe.xml: An XML configuration file that defines the object model for an adapter, along with semantic definitions for use in data analysis and management.
+- describeSchema.xsd: A file that defines what elements and attributes may appear in the `describe.xml` document. This file does not need to be included in a Management Pack but is useful for validating that the `describe.xml` file is error-free.
+
+### content (directory)
+Contains all the content included in a management pack such as: alerts, dashboards, reports, recommendations, symptoms, and traversal specs. To learn more about adding content, consult the [Adding Content](adding_content.md) guidej
+
+### resources (directory)
+The resources directory contains localization file(s), by default `resources.properties`. This file specifies the mapping of `nameKey` attribute of elements inside the `conf/describe.xml`. For more information see [Adding Localization](adding_content.md#adding-localization).
```
kjrokos, line 185: The top-level 'resources' directory has the localization file(s) for the manifest.txt file, not describe. The describe's resources folder is in the 'conf' directory next to the describe.xml file itself.

doc/java_project.md
kjrokos: Ditto for the other language-specific pages.

README.md
```diff
@@ -149,23 +150,60 @@ the creation of a new management pack project.
...
+5. Return the CollectResult.
+
+- get_adapter_definition():
+Defines the object types and attribute types (metric/property) that are present
+in a collection. Setting these object types and attribute types helps VMware Aria Operations to
+validate, process, and display the data correctly.
+
+Each method represents a single request, and it can be tested individually using `mp-test`, which is covered in the section ahead.
+
+[^1]The HTTP server does not record the state inside the executable files, so the adapter should be considered stateless.
```
kjrokos, line 200: This should be higher and reworded. We don't need to mention the HTTP server or 'executable files'. We should word the part about being stateless stronger; namely that it _is_ stateless, not that it should be _considered_ stateless. We can mention how it is run in a stateless way: That each request starts its own python subprocess which runs the appropriate method and then exits.

README.md
```diff
@@ -149,23 +150,60 @@ the creation of a new management pack project.
...
+3. Add the data into a CollectResult's objects, properties, metrics, etc
+4. Disconnect cleanly from the target (ensure this happens even if an error occurs)
+5. Return the CollectResult.
+
+- get_adapter_definition():
+Defines the object types and attribute types (metric/property) that are present
+in a collection. Setting these object types and attribute types helps VMware Aria Operations to
+validate, process, and display the data correctly.
+
+Each method represents a single request, and it can be tested individually using `mp-test`, which is covered in the section ahead.
```
kjrokos, line 198: Each method represents a single request, and it can be tested individually using `mp-test` (see section [...]).

README.md
```diff
@@ -149,23 +150,60 @@ the creation of a new management pack project.
...
+1. Read identifier values from adapter_instance that are required to connect to the target(s)
+2. Connect to the target(s), and retrieve data
+3. Add the data into a CollectResult's objects, properties, metrics, etc
+4. Disconnect cleanly from the target (ensure this happens even if an error occurs)
+5. Return the CollectResult.
+
+- get_adapter_definition():
+Defines the object types and attribute types (metric/property) that are present
+in a collection. Setting these object types and attribute types helps VMware Aria Operations to
+validate, process, and display the data correctly.
```
kjrokos, line 196: More importantly in most cases, it defines the Adapter Instance configuration parameters/credentials.

README.md
```diff
@@ -149,23 +150,60 @@ the creation of a new management pack project.
...
+ 4. If any of the above failed, return an error, otherwise pass.
+
+- get_endpoints(adapter_instance):
+This  method will be run before the 'test' method, and VMware Aria Operations will use
+the results to extract a certificate from each URL. If the certificate is not trusted by
+the VMware Aria Operations Trust Store, the user will be prompted to either accept or reject
+the certificate. If it is accepted, the certificate will be added to the AdapterInstance
+object that is passed to the 'test' and 'collect' methods. Any certificate that is
+encountered in those methods should then be validated against the certificate(s)
+in the AdapterInstance.
```
kjrokos, line 183: We should mention that this is only applicable for HTTPS endpoints. If we have a different type (e.g., a database connection), it will not work, and there is no benefit to returning anything here.

README.md
```diff
@@ -149,23 +150,60 @@ the creation of a new management pack project.
...
+[//]: # (TODO: Add this section back when we support it)
+[//]: # (8. `Select a language for the adapter`)
+[//]: # (  selected language. The template adapter collects several objects and metrics from the container that the adapter)
+[//]: # (  Once selected, the project will be generated, including a template adapter in the)
+[//]: # (  is running in, and can be used as a starting point for creating a new adapter.)
+
+For complete documentation of the `mp-init` tool including an overview of its output, see the [MP Initialization Tool Documentation](doc/mp-init.md).
+
+### Template Adapter
+The template adapter has comments throughout the code to help new users understand the process of creating their own adapter using the existing code. For additional guidance creating adapters see our [Walkthroughs](../README.md#walkthroughs) section.
```
kjrokos, line 165:
```suggestion
The template adapter has comments throughout the code to help new users understand the process of creating their own adapter using the existing code. For additional guidance creating adapters see the [Walkthroughs](../README.md#walkthroughs) section.
```

doc/mp-init.md
```diff
@@ -65,81 +65,71 @@ The _icon_ for the Management Pack generated by this project. The icon will be d
...
 
-- describe.xml: An XML configuration file that defines the object model for an adapter, along with semantic definitions for use in data analysis and management.
-- describeSchema.xsd: A file that defines what elements and attributes may appear in the `describe.xml` document. This file does not need to be included in a Management Pack but is useful for validating that the `describe.xml` file is error-free.
+### adapter_requirements.txt (file)
+This file defines all the dependencies needed by the adapter at run time. This file is copied onto the container where the adapter runs.
 
-### content (directory)
-Contains all the content included in a management pack such as: alerts, dashboards, reports, recommendations, symptoms, and traversal specs. To learn more about adding content, consult the [Adding Content](adding_content.md) guide. 
+### requirements.txt (file)
+This file defines all the dependencies needed for development of the adapter. This file does not get copied into the Dockerfile.
```
kjrokos, line 123:
```suggestion
This file defines all the dependencies needed for development of the adapter. This file does not get copied into the container image.
```

doc/mp-init.md
```diff
@@ -65,81 +65,71 @@ The _icon_ for the Management Pack generated by this project. The icon will be d
...
-and runs the process defined by the key value.
-
-### conf (directory)
-The conf directory contains configuration files used by VMware Aria Operations to create a management pack instance.   
+code, and any additional dependencies specified by the user.
 
-- describe.xml: An XML configuration file that defines the object model for an adapter, along with semantic definitions for use in data analysis and management.
-- describeSchema.xsd: A file that defines what elements and attributes may appear in the `describe.xml` document. This file does not need to be included in a Management Pack but is useful for validating that the `describe.xml` file is error-free.
+### adapter_requirements.txt (file)
+This file defines all the dependencies needed by the adapter at run time. This file is copied onto the container where the adapter runs.
```
kjrokos, line 120:
```suggestion
This file defines all the dependencies needed by the adapter at run time. This file is copied onto the container where the adapter runs and used by the default Dockerfile to install dependencies into the container image.
```

doc/mp-init.md
```diff
@@ -163,14 +153,36 @@ mark. If your editor automatically includes the byte-order mark when saving to U
...
+The template adapter has comments throughout the code to help new users understand the process of creating their own adapter using the existing code. For additional guidance creating adapters see our [Walkthroughs](../README.md#walkthroughs) section.
+
+### conf (directory)
+The conf directory contains configuration files used by VMware Aria Operations to create a management pack instance.
+
+- describe.xml: An XML configuration file that defines the object model for an adapter, along with semantic definitions for use in data analysis and management.
+- describeSchema.xsd: A file that defines what elements and attributes may appear in the `describe.xml` document. This file does not need to be included in a Management Pack but is useful for validating that the `describe.xml` file is error-free.
+
+### content (directory)
+Contains all the content included in a management pack such as: alerts, dashboards, reports, recommendations, symptoms, and traversal specs. To learn more about adding content, consult the [Adding Content](adding_content.md) guidej
```
kjrokos, line 182:
```suggestion
Contains all the content included in a management pack such as: alerts, dashboards, reports, recommendations, symptoms, and traversal specs. To learn more about adding content, consult the [Adding Content](adding_content.md) guide.
```

README.md
```diff
@@ -149,23 +150,60 @@ the creation of a new management pack project.
...
+The code inside `adapter.py` executed by an HTTP sever that runs inside container every time a test collection, or a collection is made[^1]. The HTTP server sends the information returned by the executable back to VMware Aria Operations. the template executable, `adapter.py` has four methods:
```
kjrokos, line 166: [^1] doesn't render the way I assume you want it to. Regardless, I don't think we should mention that the adapter is stateless in a footnote. I think this could use some cleaning up in general - some bits are confusing or not relevant at this point. We should talk about the architecture, but not here.

README.md
```diff
@@ -64,14 +64,15 @@ Contents
...
+* macOS 13 (Ventura)
 * Debian Linux
 * Fedora Linux
-    
+
 Other operating systems may be compatible.
 
 #### VMware Aria Operations
-The Management Packs generated by the VMware Aria Operations Integration SDK will only run on versions that supports containerized Management Packs. Currently, this is limited to on-prem installs, version 8.10 or later.
+The Management Packs generated by the VMware Aria Operations Integration SDK will only run on versions that supports containerized Management Packs. Currently, this is limited to version 8.10 or later.
```
kjrokos, line 75: This line shouldn't change. We're still on-prem only for external users.

README.md
```diff
@@ -149,23 +150,59 @@ the creation of a new management pack project.
...
+[//]: # (8. `Select a language for the adapter`)
+[//]: # (  selected language. The template adapter collects several objects and metrics from the container that the adapter)
+[//]: # (  Once selected, the project will be generated, including a template adapter in the)
+[//]: # (  is running in, and can be used as a starting point for creating a new adapter.)
+
+For complete documentation of the `mp-init` tool including an overview of its output, see the [MP Initialization Tool Documentation](doc/mp-init.md).
+
+### Template Adapter
+The template adapter has comments throughout the code to help new users understand the process of creating their own adapter using the existing code. For additional guidance creating adapters see the [Walkthroughs](../README.md#walkthroughs) section.
+The code inside `adapter.py`  has four methods:
```
kjrokos, line 166: This section starts talking about the template adapter right away without any discussion about that the template adapter is or where it's located or why we should care. Also, we should be careful about talking about things that are specific to the template adapter vs things that are generally true of any adapter. The methods below are in the latter category, so talking about them only in terms of the sample adapter is potentially misleading.

README.md
```diff
@@ -149,23 +150,62 @@ the creation of a new management pack project.
...
+[//]: # (  Once selected, the project will be generated, including a template adapter in the)
+[//]: # (  is running in, and can be used as a starting point for creating a new adapter.)
+
+For complete documentation of the `mp-init` tool including an overview of its output, see the [MP Initialization Tool Documentation](doc/mp-init.md).
+
+### Template Project
+Every new project creates a file system that has the basic project structure required to develop and build and Management Pack.
+Each file and directory is discussed in depth in the [mp-init](doc/mp-init.md) documentation. `app/adapter.py` is the entry point
+for the your adapter, and the best starting point for a new user to begin their development journey. `adapter.py` is a template
+adapter that has comments throughout its code that allow users to replace the code in the template adapter and turn it into their own adapter. The methods inside the adapter template are all required and should not be deleted; however, the code inside the methods should be substituted to generate the desired adapter. Each method is used for a different function as described below:
```
kjrokos, line 168:
```suggestion
Every new project creates a file system that has the basic project structure required to develop and build and Management Pack.
Each file and directory is discussed in depth in the [mp-init](doc/mp-init.md) documentation. `app/adapter.py` is the entry point
for the your adapter, and is the best starting point. `adapter.py` is a template
adapter that [what does it do?]. It has comments throughout its code that explain what the code in the template adapter does and how to turn it into your own adapter. The methods inside the adapter template are required and should not be deleted; however, the code inside the methods should be modified to generate the desired adapter. Each method is used for a different function as described below:
```

This reads much better than what was there before! I made some small tweaks and a section to fill in. The main thing that I noticed is you were switching between second person (e.g., "your adapter") third person (e.g., "a new user").

Also, while I appreciate the focus on new users, in many cases the same things will apply to more experienced users as well (e.g., starting with 'adapter.py'), so I'd recommend against being more specific than necessary.

README.md
```diff
@@ -149,23 +150,62 @@ the creation of a new management pack project.
...
+3. Add the data into a CollectResult's objects, properties, metrics, etc
+4. Disconnect cleanly from the target (ensure this happens even if an error occurs)
+5. Return the CollectResult.
+
+- get_adapter_definition():
+Defines the object types and attribute types (metric/property) that are present
+in a collection. Setting these object types and attribute types helps VMware Aria Operations to
+validate, process, and display the data correctly.
+
+Each method represents a single request, and it can be tested individually using `mp-test`, which is covered in the following section. The adapter is stateless,therefore the test __test__ and __collect__ methods should be self contained. For further guidance on using the template project consult the [Walkthroughs](../README.md#walkthroughs) section.
```
kjrokos, line 200: * I think this should be before the descriptions of the methods
* "self-contained" is probably not the right word, since they can certainly call other methods, libraries, files, etc. Also none of the other methods can store state either, not just the test and collect methods.

Maybe something like this:
```suggestion
Each method represents a single request, and it can be tested individually using `mp-test`, which is covered in the following section. The adapter is stateless, therefore the methods are not able to store any data for use in later method calls. For further guidance on using the template project consult the [Walkthroughs](../README.md#walkthroughs) section.
```

README.md
kjrokos: Also, 'get_adapter_definition' is optional (but recommended - if it's omitted there should be a manually-created `describe.xml` file instead)

README.md
```diff
@@ -149,23 +150,62 @@ the creation of a new management pack project.
...
+[//]: # (  Once selected, the project will be generated, including a template adapter in the)
+[//]: # (  is running in, and can be used as a starting point for creating a new adapter.)
+
+For complete documentation of the `mp-init` tool including an overview of its output, see the [MP Initialization Tool Documentation](doc/mp-init.md).
+
+### Template Project
+Every new project creates a file system that has the basic project structure required to develop and build and Management Pack.
+Each file and directory is discussed in depth in the [mp-init](doc/mp-init.md) documentation. `app/adapter.py` is the entry point
+for the your adapter, and is the best starting point. `adapter.py` is a template
+adapter that [what does it do?]. It has comments throughout its code that explain what the code in the template adapter does and how to turn it into your own adapter. The methods inside the adapter template are required and should not be deleted; however, the code inside the methods should be modified to generate the desired adapter. Each method is used for a different function as described below:
```
kjrokos, line 168: `[what does it do?]` is a placeholder - make sure you fill it in

README.md
```diff
@@ -182,7 +185,8 @@ the VMware Aria Operations Trust Store, the user will be prompted to either acce
 the certificate. If it is accepted, the certificate will be added to the AdapterInstance
 object that is passed to the 'test' and 'collect' methods. Any certificate that is
 encountered in those methods should then be validated against the certificate(s)
-in the AdapterInstance.
+in the AdapterInstance. This method will not only work against HTTPS endpoints, different types
+of endpoint will not work and there is no benefit of returning them.
```
kjrokos, line 189: We should give examples of other types of endpoints

'no benefit of returning them' this seems redundant as we just said they won't work.

README.md
```diff
@@ -195,9 +199,12 @@ Performs a collection against the target host. A typical collection will general
 - get_adapter_definition():
 Defines the object types and attribute types (metric/property) that are present
 in a collection. Setting these object types and attribute types helps VMware Aria Operations to
-validate, process, and display the data correctly.
+validate, process, and display the data correctly. More importantly, in most cases, it defines the
+Adapter Instance configuration parameters/credentials.
```
kjrokos, line 203: If it's more important it should be the first thing we talk about.

'in most cases' -> if we include this phrase, there should be some discussion about what the different cases are and why they may be relevant. 

There should be at least a sentence or two about configuration parameters and credentials and how they're used in Aria Ops

README.md
```diff
@@ -149,23 +150,69 @@ the creation of a new management pack project.
...
+5. Return the CollectResult.
+
+- get_adapter_definition():
+Defines the object types and attribute types (metric/property) that are present
+in a collection. Setting these object types and attribute types helps VMware Aria Operations to
+validate, process, and display the data correctly. More importantly, in most cases, it defines the
+Adapter Instance configuration parameters/credentials.
+
+Each method represents a single request, and it can be tested individually using `mp-test`, which is covered in the following section.
+The adapter is stateless, therefore the methods are not able to store any data for use in later method calls. For further guidance on
```
kjrokos, line 206: I think everything before 'For further guidance on...' should be in the intro section before the discussion on individual method calls

README.md
```diff
@@ -149,23 +150,68 @@ the creation of a new management pack project.
...
+
+### Template Project
+Every new project creates a file system that has the basic project structure required to develop and build a Management Pack.
+Each file and directory are discussed in depth in the [mp-init](doc/mp-init.md) documentation. `app/adapter.py` is the adapter's
+entry point and the best starting point. `adapter.py` is a template adapter that collects several objects and metrics from the
+container in which the adapter is running; use the template as a starting point for creating a new adapter. The template adapter 
+has comments throughout its code that explain what the code does and how to turn it into your adapter. The methods inside the adapter
+template are required. Modify the code inside the methods to generate the desired adapter. Each method represents a single request,
+and it can be tested individually using `mp-test`, which is covered in the following section. The adapter is stateless; therefore,
+the methods don't store any data for use in later method calls. Each method is used for a different function as described below:
```
kjrokos, line 172:
```suggestion
the adapter cannot store any data for use in later method calls. Each method is used for a different function as described below:
```

README.md
```diff
@@ -149,23 +150,68 @@ the creation of a new management pack project.
...
+
+- test(adapter_instance):
+  Performs a test connection using the information given to the adapter_instance to verify the adapter has been set up properly.
```
kjrokos, line 175:
```suggestion
  Performs a test connection using the information given to the adapter_instance to verify the adapter instance has been configured properly.
```

README.md
```diff
@@ -149,23 +150,68 @@ the creation of a new management pack project.
...
+  A typical test connection will generally consist of:
+
+     1. Read identifier values from adapter_instance that are required to connect to the target(s)
+     2. Connect to the target(s), and retrieve some sample data
+     3. Disconnect cleanly from the target (ensure this happens even if an error occurs)
+     4. If any of the above failed, return an error, otherwise pass.
```
kjrokos, line 181:
```suggestion
     3. If any of the above failed, return an error, otherwise pass.
     4. Disconnect cleanly from the target (ensure this happens even if an error occurs)
```

README.md
```diff
@@ -149,23 +150,68 @@ the creation of a new management pack project.
...
+    1. Read identifier values from adapter_instance that are required to connect to the target(s)
+    2. Connect to the target(s), and retrieve data
+    3. Add the data into a CollectResult's objects, properties, metrics, etc
+    4. Disconnect cleanly from the target (ensure this happens even if an error occurs)
+    5. Return the CollectResult.
+
+- get_adapter_definition():
+  Defines the object types and attribute types, and in most cases, it also defines the Adapter Instance configuration 
+  parameters/credentials present in a collection. Setting these helps VMware Aria Operations to validate, process, and display the 
+  data correctly.
```
kjrokos, line 204:
```suggestion
  Optional method that defines the Adapter Instance configuration (parameters and credentials used to connect to the target, and configure the management pack) present in a collection, and defines the object types and attribute types present in a collection. Setting these helps VMware Aria Operations to validate, process, and display the data correctly. If this method is omitted, a `describe.xml` file should be manually created inside the `conf` directory with the same data. Generally, this is only necessary when using advanced features of the `describe.xml` file that are not present in this method.
```

MarkArbogast:
`The Integration SDK creates Management Packs to add custom objects, data, and relationships from a endpoint into VMware Aria Operations.`

should be "an endpoint"

MarkArbogast:
`Currently, installing a Management Pack built with the integration SDK`

capitalize Integration

README.md
```diff
@@ -64,10 +64,11 @@ Contents
 The VMware Aria Operations Integration SDK has been tested in the following OSes:
```
MarkArbogast, line 64: OSes -> OSs

MarkArbogast (commented):
`run on versions that supports` ->
`run on versions that support`

MarkArbogast:
`the OS you need` -> `your OS`

MarkArbogast:
`(This script is written for the bash shell. If you use the csh or fish shells, there are alternate activate.csh and activate.fish scripts you should use instead.)`

where would one find these scripts? if we're not providing them (and I don't think we should), we should just say the script is written for the bash shell and leave it at that.

MarkArbogast:
`To exit the virtual environment, run deactivate in the virtual environment.`

why are we saying this here? if someone is following this section as instructions, we don't actually want them to deactivate their virtual environment. Regardless, our goal isn't really to teach the use of Python virtual environments. Do we need this at all?

MarkArbogast:
image

I think the `Environment → Object Browser → All Objects` is meant to be bold

MarkArbogast (approved):
changes look good

kjrokos:
Approved; please squash when you merge!

PR #110, opened 2023-04-17
Story/troubleshooting section pt 1

quirogas:
Add Troubleshooting steps for topics covering Docker CLI, Container Registry, and Adapter Issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -903,7 +903,7 @@ When everything is working as expected locally using `mp-test`, we can run
   Logs are generated and stored in the cloud proxy at were the adapter was configured at `$ALIVE_BASE/user/log/adapter/<ADAPTERKEY>_adapter3/<ADAPTER_INTERNAL_INSTANCE_ID>`.
 
   `ADAPTERKEY` should match the name of the adapter used in the `manifest.txt`, and the `ADAPTER_INTERNAL_INSTANCE_ID` should match the Internal ID
-  found in VMware Aria Operations at **Environment &rarr; Inventory &rarr; Adapter Instancer &rarr; &lt;ADAPTER_DISPLAY_NAME&gt; &rarr; &lt;ADAPTER_INSTANCE&gt;** in the rightmost column.
+  found in VMware Aria Operations at **Environment &rarr; Inventory &rarr; Adapter Instances &rarr; &lt;ADAPTER_DISPLAY_NAME&gt; &rarr; &lt;ADAPTER_INSTANCE&gt;** in the rightmost column.
   The `Internal ID` column is not displayed by default. To display the Internal ID, enable the `Internal ID` column by clicking the lower left 'column' icon and then checking the `Internal ID` box.
 
   ![highlight of the checkbox where internal id can be enbaled](doc/enable_internal_id_column.png)
@@ -964,7 +964,7 @@ When everything is working as expected locally using `mp-test`, we can run
   If the file does not exist, the system generates it after a collection/test.
 
   `ADAPTERKEY` should match the name of the adapter used in the `manifest.txt`, and the `ADAPTER_INTERNAL_INSTANCE_ID` should match the Internal ID
-  found in VMware Aria Operations at **Environment &rarr; Inventory &rarr; Adapter Instancer &rarr; &lt;ADAPTER_DISPLAY_NAME&gt; &rarr; &lt;ADAPTER_INSTANCE&gt;** in the rightmost column.
+  found in VMware Aria Operations at **Environment &rarr; Inventory &rarr; Adapter Instances &rarr; &lt;ADAPTER_DISPLAY_NAME&gt; &rarr; &lt;ADAPTER_INSTANCE&gt;** in the rightmost column.
   To ensure Internal ID is displayed ensure the Internal ID box is enabled by clicking in the bottom left icon and then the checkbox.
 
   ![highlight of the checkbox where internal id can be enbaled](doc/enable_internal_id_column.png)


README.md
```diff
@@ -258,7 +258,49 @@ TODO
 
 
 ## Troubleshooting
-TODO
+
+If you encounter any issues while using the VMware Aria Operations Integration SDK, you can refer to the troubleshooting guides for each tool:
+
+- [mp-test Troubleshooting Guide](doc/mp-test.md#troubleshooting)
+- [mp-init Troubleshooting Guide](doc/mp-init.md#troubleshooting)
+- [mp-build Troubleshooting Guide](doc/mp-build.md#troubleshooting)
```
kjrokos, line 266: What we're really looking for is something like
```
### Problem/Question

Text describing the solution
```

So things like this are OK as a catch-all at the bottom, but maybe better would be to duplicate some of the more common issues directly. For example:
```
### I'm getting 'SyntaxError: invalid syntax' when I try to run the SDK tools

<Text from mp-init troubleshooting>
```

README.md
```diff
@@ -242,23 +242,67 @@ For complete documentation of the `mp-build` tool see the [MP Build Tool Documen
...
+
+### Registry Issues:
+TODO:
+### Dockerfile Issues:
+- Error when modifying Dockerfile
+TODO:
+### Adapter Code:
+Logs are generated and stored in the `logs` directory whenever the adapter runs locally using' mp-test'. There are three types of log files: adapter, server, and test logs. Each log file is prepended with the type of log file followed by a number representing rollover, if there was any, and the .log file extension. Server logs come from the HTTP server that runs inside the container. Server logs can't be modified since the server code comes packaged inside the base-adapter Python image. Adapter logs are all the logs generated by adapter code (e.g., the test() method or the collect() methods inside `app/adapter.py`). Finally, test logs come from the logs generated by `mp-test`. Server and Adapter log levels are set inside a configuration file (logs/loglevels.cfg).- Setting log level for Adapter:
```
kjrokos, line 282: The text here is good -- but let's make sure it fits the 'Problem/Question - Solution' format. This could easily be fixed by adding some headers. You're actually grouping a bunch of things together! This could easily be broken up into sections, like this:
```
### Where does my adapter log to?
(also, make sure we tell them how to get to the log directory on the Cloud Proxy)

### What are the different log files?

### How do I log in my adapter code?

### How do I change the log level?
```
None of the solution text needs to be more than a sentence or two - I think you can just about copy/paste sentences you've already written above and put them in the appropriate section (double check that they don't need any light editing of course)

README.md
```diff
@@ -242,23 +242,67 @@ For complete documentation of the `mp-build` tool see the [MP Build Tool Documen
...
+
+If you encounter any issues while using the VMware Aria Operations Integration SDK, you can refer to the troubleshooting guides for each tool:
+
+- [mp-test Troubleshooting Guide](doc/mp-test.md#troubleshooting)
+- [mp-init Troubleshooting Guide](doc/mp-init.md#troubleshooting)
+- [mp-build Troubleshooting Guide](doc/mp-build.md#troubleshooting)
+
+### Docker Issues
+
+If you encounter any issues with Docker while using the VMware Aria Operations Integration SDK, you can refer to the following guides:
```
kjrokos, line 270: This says 'any issues', but the two guides are just about getting Docker running. Like before, the text is generally pretty good but let's make sure the heading is a specific problem/question that matches the text.

README.md
```diff
@@ -300,9 +300,156 @@ additional data, using the MySQL Management Pack as an example.</summary>
...
+<h3> Cannot connect to docker daemon (Windows)?</h3>
+</summary>
+If you're having trouble getting Docker to run on your system, you can refer to the Docker documentation for instructions on how to start Docker on [macOS](https://docs.docker.com/docker-for-mac/install/), [Linux](https://docs.docker.com/desktop/install/debian/#launch-docker-desktop), and [Windows 10 and 11](https://docs.docker.com/desktop/install/windows-install/#start-docker-desktop).
+</details>
+
+<details>
+<summary>
+<h3> Permission denied while trying to connect to the Docker daemon?</h3>
+</summary>
+If you're having trouble with permissions on a Windows system, you can refer to the Docker documentation for instructions on how to [Understand permission requirements for Windows](https://docs.docker.com/desktop/windows/permission-requirements/).
```
kjrokos, line 328: Are the titles here switched? The first mentions windows, but the text talks about Mac and linux as well, whereas the second only talks about windows.

Do we have records of the actual errors? Those would be good to use for titles if we have them.

README.md
```diff
@@ -300,9 +300,156 @@ additional data, using the MySQL Management Pack as an example.</summary>
...
+<h3> Where are the adapter logs stored locally? </h3>
+</summary>
+Logs are generated and stored in the `logs` directory whenever the adapter runs locally using' mp-test'
+</details>
+
+<details>
+<summary>
+<h3> Where the adapter logs stored VMware Aria Operations?</h3>
+</summary>
+Logs are generated and stored in the cloud proxy at `$ALIVE_BASE/user/log/adapter/<ADAPTERNAME>_adapter3/<ADAPTER_INTERNAL_INSTANCE_ID>`. ADAPTERNAME should match the name of the adapter used in the manifest.txt, and the ADAPTER_INTERNAL_INSTANCE_ID should match the Internal ID found in VMware Aria Operations at Environment>Inventory>Adapter Instances>My Adapter Adapter Instance>Instance** in the rightmost column.
```
kjrokos, line 361: A screenshot showing the internal id might be useful. Also I believe that column isn't shown by default.

README.md
```diff
@@ -300,9 +300,156 @@ additional data, using the MySQL Management Pack as an example.</summary>
...
+Contains all logs related to `mp-test`
+
+- build.log
+Contains all logs related to `mp-build`
+
+- validation.log
+Contains all logs related to the validation of the collection result(s).
+
+  Adapter logs are all the logs generated by adapter code (e.g., the test() method or the collect() methods inside
+`app/adapter.py`). Finally, test logs come from the logs generated by `mp-test`.
```
kjrokos, line 388: This should go in the `adapter.log` section

README.md
```diff
@@ -300,9 +300,156 @@ additional data, using the MySQL Management Pack as an example.</summary>
...
+Contains all logs related to the adapter
+
+- test.log
+Contains all logs related to `mp-test`
+
+- build.log
+Contains all logs related to `mp-build`
+
+- validation.log
+Contains all logs related to the validation of the collection result(s).
```
kjrokos, line 385: May be somewhat obvious, but it could be useful to mention that these are only found locally

README.md
```diff
@@ -300,9 +300,156 @@ additional data, using the MySQL Management Pack as an example.</summary>
...
+If the file does not exist, it will be generated after a collection/test collection.
+</details>
+
+
+<details>
+<summary>
+<h3> 500 INTERNAL SERVER ERROR:</h3>
+</summary>
+
+Internal sever can happen for various reasons; however, the most common reason is due to an unhandled exception or syntax errors in
```
kjrokos, line 428:
```suggestion
Internal server errors can happen for various reasons; however, the most common reason is due to an unhandled exception or syntax errors in
```

README.md
```diff
@@ -300,9 +300,205 @@ additional data, using the MySQL Management Pack as an example.</summary>
...
+
+  1. [Log into to your registry using aws CLI](https://docs.aws.amazon.com/AmazonECR/latest/userguide/getting-started-cli.html#cli-authenticate-registry)
+  2. [Create a repository](https://docs.aws.amazon.com/AmazonECR/latest/userguide/getting-started-cli.html#cli-create-repository:~:text=your%20default%20registry-,Step%203%3A%20Create%20a%20repository,-Step%204%3A%20Push)
+  3. Run mp-build and use the registry tag when propted about it (usually looks like `aws_account_id.dkr.ecr.region.amazonaws.com/hello-repository`)
+</details>
+
+<details>
+  <summary><h3>How can I set up a Dockerhub container registry for my project?</h3></summary>
+
+  Docker CLI recommends using a token when using docker hub instead of using your login password, so users should authenticate their dockerhub account before runing `mp-build`.
```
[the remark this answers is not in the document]
MarkArbogast: also: `runing` -> `running`

README.md
```diff
@@ -835,8 +835,205 @@ When everything is working as expected locally using `mp-test`, we can run
...
+
+<details>
+  <summary><h3> How do I change the log level (Server and Adapter)?</h3></summary>
+
+  Server and Adapter log levels are set inside the `loglevels.cfg`; this file is located in `logs/loglevels.cfg` locally and `$ALIVE_BASE/user/log/adapters/<ADAPTERKEY>_adapter3/<ADAPTER_INTERNAL_INSTANCE_ID>/loglevels.cfg` in the cloud proxy.
+  If the file does not exist, it will be generated after a collection/test collection.
+
+  ADAPTERKEY should match the name of the adapter used in the manifest.txt, and the ADAPTER_INTERNAL_INSTANCE_ID should match the Internal ID
+  found in VMware Aria Operations at Environment>Inventory>Adapter Instances>My Adapter Adapter Instance>Instance** in the rightmost column.
+  To ensure Internal ID is displayed ensure the Internal ID box is enabled by clicking in the bottom left icon and then the checkbox.
```
kjrokos, line 976: This part is a little hard to read when rendered in markdown.

Some things that might help:
* put the all-caps variables backpacks
* Bold the navigation hierarchy stand out a bit more (maybe you meant to do that, I see `**` on one side), and separate the levels better (maybe use an actual arrow (` &rarr; `) instead of just `>`?)

README.md
```diff
@@ -300,9 +300,205 @@ additional data, using the MySQL Management Pack as an example.</summary>
...
+
+<details>
+  <summary><h3>How can I set up an AWS container registry for my project?</h3></summary>
+
+  AWS container registries use `aws` CLI to authenticate, so users should authenticate to their AWS container registry and create a repository before
+  running `mp-build`.
+
+  1. [Log into to your registry using aws CLI](https://docs.aws.amazon.com/AmazonECR/latest/userguide/getting-started-cli.html#cli-authenticate-registry)
+  2. [Create a repository](https://docs.aws.amazon.com/AmazonECR/latest/userguide/getting-started-cli.html#cli-create-repository:~:text=your%20default%20registry-,Step%203%3A%20Create%20a%20repository,-Step%204%3A%20Push)
+  3. Run mp-build and use the registry tag when propted about it (usually looks like `aws_account_id.dkr.ecr.region.amazonaws.com/hello-repository`)
```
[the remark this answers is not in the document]
kjrokos: I think in most cases we use backpacks for `mp-build` and friends. In fact, that's what we're using just above on line 866

MarkArbogast (approved):
Looks good!

README.md
```diff
@@ -835,8 +835,205 @@ When everything is working as expected locally using `mp-test`, we can run
...
+  <summary><h3> How do I change the log level (Server and Adapter)?</h3></summary>
+
+  Server and Adapter log levels are set inside the `loglevels.cfg`; this file is located in `logs/loglevels.cfg` locally and `$ALIVE_BASE/user/log/adapters/<ADAPTERKEY>_adapter3/<ADAPTER_INTERNAL_INSTANCE_ID>/loglevels.cfg` in the cloud proxy.
+  If the file does not exist, it will be generated after a collection/test collection.
+
+  ADAPTERKEY should match the name of the adapter used in the manifest.txt, and the ADAPTER_INTERNAL_INSTANCE_ID should match the Internal ID
+  found in VMware Aria Operations at Environment>Inventory>Adapter Instances>My Adapter Adapter Instance>Instance** in the rightmost column.
+  To ensure Internal ID is displayed ensure the Internal ID box is enabled by clicking in the bottom left icon and then the checkbox.
+
+  ![highlight of the checkbox where internal id can be enbaled](doc/enable_internal_id_column.png)
```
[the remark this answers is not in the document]
kjrokos: I think you highlighted the wrong one

README.md
```diff
@@ -835,8 +835,212 @@ When everything is working as expected locally using `mp-test`, we can run
...
+ Running `mp-test` or `mp-build` generates logs.  Logs are generated and stored in the `logs` directory.
+</details>
+<details>
+  <summary><h3> Where are the adapter logs stored in VMware Aria Operations?</h3></summary>
+
+  Logs are generated and stored in the cloud proxy at were the adapter was configured at `$ALIVE_BASE/user/log/adapter/<ADAPTERKEY>_adapter3/<ADAPTER_INTERNAL_INSTANCE_ID>`.
+
+  `ADAPTERKEY` should match the name of the adapter used in the `manifest.txt`, and the `ADAPTER_INTERNAL_INSTANCE_ID` should match the Internal ID
+  found in VMware Aria Operations at **Environment &rarr; Inventory &rarr; Adapter Instances &rarr; My Adapter Adapter Instance &rarr; Instance** in the rightmost column.
+  To ensure Internal ID is displayed ensure the Internal ID box is enabled by clicking in the bottom left icon and then the checkbox.
```
kjrokos, line 917:
```suggestion
  The `Internal ID` is not displayed by default. To display the Internal ID, enable the **Internal ID** column by clicking the lower left 'column' icon and then checking the Internal ID box.
```

README.md
```diff
@@ -300,7 +300,7 @@ This guide assumes you have already set up the SDK and know how to create a new
 It walks you through the steps necessary to extend an existing Management Pack to add
 additional data, using the MySQL Management Pack as an example.</summary>
 
-
+--
```
kjrokos, line 303: Seems like some stray characters here

README.md
```diff
@@ -835,8 +835,212 @@ When everything is working as expected locally using `mp-test`, we can run
...
-## Troubleshooting
-TODO
+# Troubleshooting
+
+If you encounter any issues while using the VMware Aria Operations Integration SDK tools, you
+can refer to the troubleshooting guides for each tool:
+
+- [mp-test Troubleshooting Guide](doc/mp-test.md#troubleshooting)
+- [mp-init Troubleshooting Guide](doc/mp-init.md#troubleshooting)
+- [mp-build Troubleshooting Guide](doc/mp-build.md#troubleshooting)
```
kjrokos, line 845: Lets remove this or at least move it to the bottom of the page. The items that are in each of those troubleshooting guides generally aren't important enough to be at the top of the page. If there are any you feel that are, let's copy them here as well.

README.md
kjrokos: This looks much better. Maybe so something to set apart **My Adapter Adapter Instance** since the others are all literal, but that is a placeholder value?

kjrokos (approved):

kjrokos:
Approved. Please squash changes!

PR #113, opened 2023-04-19
Documentation/sample mysql extension mp

kjrokos:
Adds the complete MySQL Extension MP to the SDK, and links to it from the README.


MarkArbogast:
Looks great! I was able to get it working. If not already there, we could add a note that when using this sample, setting the Suite API connection information is required. It makes sense that you'd need to, but would be nice to have a reminder.

Approved pending consideration of this idea.

MarkArbogast (approved):

PR #122, opened 2023-04-24
Add key, label, and order for enums values

quirogas:
Modify the API schema file to support the dictionary of values containing order, enum key, and enum label.


kjrokos (commented):
`enum_values` still needs to be an array so we can have multiple enum options, we're just switching out the string literal for the object inside of it. 

The object looks good, but let's switch `order` to `display_order` for consistency.

Can you make sure you pre-commit checks are working? There's a few minor formatting issues that pre-commit should catch.

kjrokos (approved):

PR #127, opened 2023-04-27
Bug/modify definition

quirogas:
- Update definitions

### DIFF
--- a/lib/python/src/aria/ops/definition/credential_type.py
+++ b/lib/python/src/aria/ops/definition/credential_type.py
@@ -139,8 +139,6 @@ class CredentialEnumParameter(CredentialParameter):
         ):
             self.values.append((default, default))
 
-        # set the default value to the value of the label if they are diferent, otherwise set to the passed default
-
     def to_json(self) -> dict:
         return super().to_json() | {
             "type": "string",


lib/python/src/aria/ops/definition/credential_type.py
```diff
@@ -122,24 +122,28 @@ class CredentialEnumParameter(CredentialParameter):
     def __init__(
         self,
         key: str,
-        values: list[str],
+        values: list[Union[str, tuple[str, str]]],
         label: Optional[str] = None,
-        default: Optional[str] = None,
+        default: Optional[Union[str, tuple[str, str]]] = None,
```
kjrokos, line 127: no reason to change this. If someone wants a label for the default, they should add it to 'values'

lib/python/src/aria/ops/definition/credential_type.py
```diff
@@ -122,24 +122,28 @@ class CredentialEnumParameter(CredentialParameter):
...
+            elif default not in [v[0] if isinstance(v, tuple) else v for v in self.values]:
+                self.values.append((default, default))
 
     def to_json(self) -> dict:
         return super().to_json() | {
             "type": "string",
             "default": self.default,
             "enum": True,
-            "enum_values": [str(value) for value in self.values],
+            "enum_values": [{"key": str(value[0]) if isinstance(value, tuple) else value, "label": str(value[1]) if isinstance(value, tuple) else value, "display_order": display_order} for display_order, value in enumerate(self.values)]
```
kjrokos, line 146: This looks correct but is hard to read.

lib/python/src/aria/ops/definition/parameter.py
```diff
@@ -121,10 +121,10 @@ class EnumParameter(Parameter):
     def __init__(
         self,
         key: str,
-        values: list[str],
+        values: list[Union[str, tuple[str, str]]],# [{key: True, label: True }, {key: False, label: False }]
```
kjrokos, line 124: Stray comment

lib/python/src/aria/ops/definition/parameter.py
```diff
@@ -147,14 +148,25 @@ def __init__(
...
+        if (
+            default not in [v[0] if isinstance(v, tuple) else v for v in self.values]
+            and default is not None
+        ):
+            self.values.append((default, default))
 
     def to_json(self) -> dict:
         return super().to_json() | {
-            "type": "string",
+            "type": "dict",
```
kjrokos, line 161: The 'type' here is what type of value the identifier holds, not the type of the json. It should remain 'string', because enums return a string rather than a numeric value

lib/python/tests/definition/test_object_type_definition.py
```diff
@@ -95,11 +95,15 @@ def test_define_enum_identifier() -> None:
...
+            {"key": "B", "label": "B", "display_order": 0},
+            {"key": "C", "label": "C", "display_order": 1},
+            {"key": "E", "label": "E", "display_order": 2},
+        ],
         "description": None,
         "default": "E",
         "display_order": 0,
-        "type": "string",
+        "type": "dict",
     }
```
kjrokos, line 109: Would be good to have some additional tests that set the label to something other than the key.

lib/python/src/aria/ops/definition/credential_type.py
kjrokos: What should generally happen is this:
```
enum_values: [("key1", "label1"), ("key2", "label2"), ("key3", "label3")]
default: "key2"
```
In which the default will have the label "label2", but we don't specify it in the 'default' section.

We'll also allow something like this
```
enum_values: [("key1", "label1"), ("key2", "label2"), ("key3", "label3")]
default: "key4"
```
In which the default will have the label "key4". It's possible we may want to not allow this case, but it would technically be a breaking change.

We want to not get into this situation:
``
enum_values: [("key1", "label1"), ("key2", "label2"), ("key3", "label3")]
default: ("key2", "default_label")
```
In the above, what is the label for 'key2' - "label2" or "default_label"? We can only set one.

File: lib/python/src/aria/ops/definition/credential_type.py [hunk 6]
```diff
@@ -122,24 +123,40 @@ class CredentialEnumParameter(CredentialParameter):
...
+            default not in [v[0] if isinstance(v, tuple) else v for v in self.values]
+            and default is not None
+        ):
+            self.values.append((default, default))
+
+        # set the default value to the value of the label if they are diferent, otherwise set to the passed default
+        self.default = next(
+            (v[1] for v in self.values if isinstance(v, tuple) and v[0] == default),
+            default,
+        )
```
kjrokos, line 145: Why are we setting the default to its label? The key should be what we want.

lib/python/src/aria/ops/definition/parameter.py
```diff
@@ -15,7 +16,7 @@ def __init__(
         key: str,
         label: Optional[str] = None,
         description: Optional[str] = None,
-        default: Optional[str | int] = None,
+        default: Optional[Union[str, int, tuple[str, str]]] = None,
```
kjrokos, line 19: Type is wrong here - default should never be a tuple

lib/python/src/aria/ops/definition/credential_type.py
```diff
@@ -139,10 +140,6 @@ def __init__(
             self.values.append((default, default))
 
         # set the default value to the value of the label if they are diferent, otherwise set to the passed default
```
kjrokos: Comment is no longer applicable

kjrokos (approved):

PR #128, opened 2023-04-28
Add version to AdapterDefinition endpoint

quirogas:
- Add Server version to Adapter Definition call

### DIFF
--- a/lib/python/src/aria/ops/definition/adapter_definition.py
+++ b/lib/python/src/aria/ops/definition/adapter_definition.py
@@ -74,7 +74,6 @@ class AdapterDefinition(GroupType):  # type: ignore
             "adapter_key": self.key,
             "adapter_label": self.label,
             "describe_version": self.version,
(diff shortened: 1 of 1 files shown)


vmware_aria_operations_integration_sdk/api/integration-sdk-definition-endpoint.json
```diff
@@ -322,6 +322,9 @@
         "type": "object",
         "description": "Definition of the object model returned in CollectResult",
         "properties": {
+          "version":{
```
kjrokos, line 325: "version" is ambiguous, especially since it's inside an 'AdapterDefinition' object. We should clarify that it is the version of the adapter definition schema, rather than the adapter definition itself (this is especially true because the describe.xml file is also versioned, but it _is_ the for versioning the adapter definition itself).

kjrokos (approved):

PR #167, opened 2023-05-03
Bug/read updated definitions

quirogas:
- Read new definitions 
- generate describe.xml based on the schema version

### DIFF
--- a/vmware_aria_operations_integration_sdk/describe.py
+++ b/vmware_aria_operations_integration_sdk/describe.py
@@ -388,7 +388,7 @@ def add_enum_values(
     parent: Element, identifier_json: Dict, names: _Names, schema_version: int
 ) -> None:
     if "enum_values" in identifier_json:
-        if schema_version == 1:
+        if schema_version >= 1:
             _add_enum_values_v1(parent, identifier_json, names)
         else:
             _add_enum_values_v0(parent, identifier_json, names)


vmware_aria_operations_integration_sdk/describe.py
```diff
@@ -368,24 +380,48 @@ def add_identifier(parent: Element, identifier_json: Dict, names: _Names) -> Ele
...
-                "enum",
-                attrib={
-                    "value": str(value),
-                    "default": str(
-                        value == identifier_json.get("default", False)
-                    ).lower(),
-                },
-                nsmap=ns_map,
-            )
+        if schema_version == "1.0.0":
```
kjrokos, line 391: If we have to start supporting a bunch of versions, it might be easier to just use integers - otherwise we'll have to parse this version string in order to do logical comparisons.

vmware_aria_operations_integration_sdk/container_versions.json
```diff
@@ -2,7 +2,7 @@
     "base_image": {
         "language": "Python",
         "path": "base-python-adapter",
-        "version": "0.11.0"
+        "version": "0.12.0"
```
kjrokos, line 5: Have you tested to see if the old schema will work with this server? We will definitely want to do that, as if it does not we need to make sure we're very clear that this is a breaking change.

That said, I _expect_ that the old schema will work - I just don't know with certainty.

vmware_aria_operations_integration_sdk/describe.py
```diff
@@ -380,15 +380,15 @@ def add_identifier(
...
     return identifier_xml
 
 
 def add_enum_values(
-    parent: Element, identifier_json: Dict, schema_version: str, names: _Names
+    parent: Element, identifier_json: Dict, names: _Names, schema_version: int
 ) -> None:
     if "enum_values" in identifier_json:
-        if schema_version == "1.0.0":
+        if schema_version == 1:
```
kjrokos, line 391: For future compatibility, this should be `>= 1`

kjrokos: If we make another change to enums in the schema, then we can modify then, but if we make a change to an unrelated part of the schema, then we don't have to go through all the existing places where we're checking the schema and modify those too.

kjrokos (approved):

PR #170, opened 2023-05-03
Add support for labels of enum values

### LINKED ISSUE #84: Support localization (nameKeys) on Parameter Enums

Currently enum values for a enum parameter do not have a label - the key is displayed. In most other instances we're allowed to use a label, which supports localization and decoupling implementation from UI code. We didn't support this previously because it's not allowed by the describe.xsd spec, but in this case the spec is wrong.

This isn't in the describe.xsd, but it does work. We should amend the describe.xsd and support the feature. To do this we may need to switch enum json representation from a raw string to a object with key/label. If we also add displayOrder, this can address #83 as well.

_No response_

### Additional context

_No response_

quirogas:
* Add labeled enum values support
* Add `schema_version` to adapter definition endpoint to track the schema version used by the Adapter 
* Update Credential and Identifier enum type to support labels and display orders for each enum value 

resolves #84

### DIFF
--- a/images/base-python-adapter/CHANGELOG.md
+++ b/images/base-python-adapter/CHANGELOG.md
@@ -1,5 +1,5 @@
-VMware Aria Operations Integration Base Python Adapter 
+VMware Aria Operations Integration Base Python Adapter
 ----------------------------------------------
-## Unreleased
-* Add `schema_version` to adapter definition endpoint to track the schema version used by the Adapter 
-* Update Credential and Identifier enum type to support labels and display orders for each enum value 
+## 0.12.0 (05-03-2023)
+* Add `schema_version` to adapter definition endpoint to track the schema version used by the Adapter
+* Update Credential and Identifier enum type to support labels and display orders for each enum value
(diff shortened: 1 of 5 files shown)


vmware_aria_operations_integration_sdk/api/integration-sdk-definition-endpoint.json
```diff
@@ -296,6 +322,9 @@
         "type": "object",
         "description": "Definition of the object model returned in CollectResult",
         "properties": {
+          "schema_version":{
+            "type":"int"
```
kjrokos, line 326: I thought it was strange that the generated code was using a special 'ModelInt' type for this - Now that I see what we used for `display_order` I think it's because we want `integer` for the type:
```
schema_version:
  type: integer           
  format: int32
```

images/base-python-adapter/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+VMware Aria Operations Integration Base Python Adapter 
+----------------------------------------------
+## Unreleased
```
kjrokos, line 3: Did we release 0.12.0? If so we should mark that here

kjrokos (approved):

PR #176, opened 2023-05-10
Release SDK version 0.5.1

### LINKED ISSUE #126: Warnings on template adapter
### Describe the bug

When testing the template MP with no changes, there are warnings

### Reproduction steps

1. Run mp-init
2. Run mp-test collect
3. Collect result from mp-test has warnings

### Expected behavior

There should be no warnings when running the samples or templates

### Additional context

The warnings are occurring because the object type keys do not match between collections and definitions.

kjrokos:
* Updates docker module to 6.1.1, which resolves urllib3 2.0.0/requests 2.29 incompatibility
* Updates sdk lib to version 0.7.3, for labeled enum value support
* Recommend using `pipx` over `pip` in installation guide to decrease likelihood of dependency conflict in the global python environment
* Adds icon folders to 'conf' directory when generating a new project
* Updates adapter template to current best practices

Resolves #126


MarkArbogast:
Looks good but CI failed on macos. Approved pending fix for macos CI.

PR #179, opened 2023-05-15
Add Python documentation

quirogas:
Add Python docs for the SDK library:
image


kjrokos:
In your screenshot, it looks like the default argument for `send_results` is being evaluated - the default _should_ be `sys.argv[-1]`, but it looks like it's being evaluated to the last argument of the `pdoc` command. Any way to avoid this?

kjrokos (approved):

PR #183, opened 2023-05-17
Update documentation to highlight XML generation logic

quirogas:
- Remove `doc/python-integration-module.md` (to favor the Python documentation)
- Update `doc/adding_to_an_adapter.md`
   - Add an explanation about describe.xml generation and show side-by-side examples
   - Explain methods to add relationships between objects(updated).
   - Explain Credential Kind and its usage (updated).

### DIFF
--- a/doc/references.md
+++ b/doc/references.md
@@ -24,22 +24,4 @@ Contents
   * commands.cfg
 
 * [VMware Aria Operations Integration SDK Library](https://pypi.org/project/vmware-aria-operations-integration-sdk-lib/)
-  * [Adapter Instance](../lib/python/doc/aria/ops/adapter_instance.html)
-  * [Adapter Logging](../lib/python/doc/aria/ops/adapter_logging.html)
-  * [Data](../lib/python/doc/aria/ops/data.html)
-  * Definition
-     * [Adapter Definition](../lib/python/doc/aria/ops/definition/adapter_definition.html)
-     * [Assertions](../lib/python/doc/aria/ops/definition/assertions.html)
-     * [Attribute](../lib/python/doc/aria/ops/definition/attribute.html)
-     * [Credential Type](../lib/python/doc/aria/ops/definition/credential_type.html)
-     * [Exceptions](../lib/python/doc/aria/ops/definition/exceptions.html)
-     * [Group](../lib/python/doc/aria/ops/definition/group.html)
-     * [Object Type](../lib/python/doc/aria/ops/definition/object_type.html)
-     * [Parameter](../lib/python/doc/aria/ops/definition/parameter.html)
-     * [Units](../lib/python/doc/aria/ops/definition/units.html)
-  * [Event](../lib/python/doc/aria/ops/event.html)
-  * [Object](../lib/python/doc/aria/ops/object.html)
-  * [Pipe Utils](../lib/python/doc/aria/ops/pipe_utils.html)
-  * [Result](../lib/python/doc/aria/ops/result.html)
-  * [Suite API Client](../lib/python/doc/aria/ops/suite_api_client.html)
-  * [Timer](../lib/python/doc/aria/ops/timer.html)
+  


doc/adding_to_an_adapter.md
```diff
@@ -1,120 +1,237 @@
...
+>     definition = AdapterDefinition("MyAdapter", "My Adapter")
+>
+>     definition.define_string_parameter(
+>         "instance",
+>         label="Instance",
+>         required=True,
+>     )
+>     definition.define_enum_parameter(
+>         "ssl_mode",
+>         values= ["Disable"],
```
kjrokos, line 71: I'd suggest we should always add all the enum values here, rather than relying on the `default` value being automatically added:
```suggestion
>         values=["Disable", "Require"],
```

doc/adding_to_an_adapter.md
```diff
@@ -1,120 +1,237 @@
...
+> The `AdapterDefinition` will generate a describe.xml equivalent to the one below
+> ### describe.xml
+> ```xml
+> <AdapterKind xmlns="http://schemas.vmware.com/vcops/schema" key="My Adapter" nameKey="1" version="1">
+>   <ResourceKinds>
+>     <ResourceKind key="MyAdapter_adapter_instance" nameKey="5" type="7" >
+>         <ResourceIdentifier dispOrder="1" key="instance" nameKey="6" required="true" type="string" identType="1"/>
+>         <ResourceIdentifier dispOrder="2" key="ssl_mode" nameKey="7" required="true" type="string" identType="2" enum="true">
+>             <enum default="true" value="Disable" />
+>             <enum default="false" value="Require" />
```
kjrokos, line 91: The default values are different between this XML code and the python code

doc/adding_to_an_adapter.md
```diff
@@ -1,120 +1,237 @@
...
+>   <!-- ... -->
+> </AdapterKind>
+> ```
 For more information about the supported elements and attributes, see the [describe.xml documentation](../vmware_aria_operations_integration_sdk/adapter_template/describeSchema.xsd).
+For more information about the Python module visit [VMware Aria Operations Integration SDK Library](https://pypi.org/project/vmware-aria-operations-integration-sdk-lib/).
+
+In the `describe.xml` adapter instance identifiers can have an `identType` of `1` or `2`. A type of `1` means the
+identifier will be used for determining uniqueness, and will show up by default on the configuration page. If the type
+is `2`, the identifier is _non-identifying_, and will show up under the 'advanced' section of the configuration page.
+In Python, the [Parameter](../lib/python/doc/aria/ops/definition/parameter.html) object has an `adavanced` attribute that determines
```
kjrokos, line 105: Typo in 'advanced' here. Would be good to run a spell check.

doc/adding_to_an_adapter.md
```diff
@@ -1,120 +1,237 @@
...
-After the account has been created, credential fields will be available in the input to the `collect`, `test_connection`, and `get_endpoint_urls` methods. (See [Creating an Adapter Instance](#creating-an-adapter-instance).)
-
-Using the [Python VMware Aria Operations Integration Module](python-integration-module.md), the credential is available in the input to the `collect`, `test_connection`, and `get_endpoint_urls` methods. (See [Creating an Adapter Instance](#creating-an-adapter-instance).)
+> ![Adding a credential](adding_a_credential.png)
+>
+> Adding a credential to an adapter instance with the `username` and `password` fields as defined above. 'Credential name'
+> is always added (by VMware Aria Operations), and allows for credentials to be reused between adapter instances.
+
+Both `describe.xml` and [AdapterDefinition](../lib/python/doc/aria/ops/definition/adapter_definition.html) allow for the
+use of multiple credential.
```
kjrokos, line 178:
```suggestion
use of multiple credential types.
```

doc/adding_to_an_adapter.md
```diff
@@ -1,120 +1,237 @@
...
+> def get_adapter_definition() -> AdapterDefinition:
+>     definition = AdapterDefinition("MyAdapter", "My Adapter")
+>
+>     credential = definition.define_credential_type("my_credential_type", "Credential")
+>     credential.define_string_parameter("user_name", "User Name")
+>     credential.define_password_parameter("user_password", "Password", required=False)
+>     # Second set of credentials
+>     second_credential = definition.define_credential_type("my_second_credential_type", "Second Credential")
+>     second_credential.define_string_parameter("second_user_name", "Second User Name")
+>     second_credential.define_password_parameter("second_user_password", "Second Password", required=False)
```
kjrokos, line 191: I think if we do an example we should make sure it makes sense, so a user will have an example of _why_ they might need two credential kinds -- there's no point in having two credential kinds if they both have the same two pieces of information.

doc/adding_to_an_adapter.md
```diff
@@ -1,120 +1,237 @@
...
+>> In the describe.xml credential kind keys are separated by a comma.
+
+Once the credential is defined in the [object model](#defining-an-adapter-and-adapter-instance-in-the-object-model), it can be used in the adapter code.
+> Note: If there are any existing connections used by the [`mp-test`](mp-test.md) tool before the credential was created
+> or updated, these will need to be deleted or updated.
+
+Once an adapter instance is defined, any credential defined in the [object model](#defining-an-adapter-and-adapter-instance-in-the-object-model)
+will be prompted to the user when creating an account in VMware Aria Operations on the `Data Sources` &rarr; `Integrations`
+page. After the account has been created, credential fields will be available `AdapterInstance` object passed to the
+`collect`, `test`, and `get_endpoints` methods.(See [Creating an Adapter Instance](#creating-an-adapter-instance).)
```
kjrokos, line 221: May want to mention `get_credential_type` for determining which type of credential the user has supplied, if there are more than one

doc/adding_to_an_adapter.md
```diff
@@ -125,40 +242,86 @@ password = adapter_instance.get_credential_value("password")
...
-the [VMware Aria Operations Collector Framework OpenAPI Document](../vmware_aria_operations_integration_sdk/api/vmware-aria-operations-collector-fwk2.json)
-.
-
-## Adding an Object Type
-An object type is a class of objects (resources) that share the same set of metrics, properties, and identifiers. For example an adapter might have a 'Database' object kind, and when an adapter instance is created and connects to an application, several 'database' objects are created representing distinct databases in the application. To create a new object type, add a `ResourceKind` element inside `AdapterKind/ResourceKinds` in the `conf/describe.xml` file.
-A `key` attribute is required, and must be unique among other object types within the `describe.xml` file. 
+For other languages, or using Python without the VMware Aria Operations Integration SDK Library module, JSON representing
+the adapter instance is sent to a named pipe. The second-to-last argument the adapter is invoked with will always be the
+filename of the named pipe.
+
```
kjrokos, line 248: Any reason you removed the reference to the API document?

doc/adding_to_an_adapter.md
```diff
@@ -1,120 +1,237 @@
...
-After the account has been created, configuration fields will be available in the input to the `collect`, `test_connection`, and `get_endpoint_urls` methods. (See [Creating an Adapter Instance](#creating-an-adapter-instance).)
+Once an adapter instance is defined, any configuration fields (`ResourceIdentifiers` element) and credentials (`CredentialKind` element)
+will be prompted to the user when creating an account in VMware Aria Operations on the `Data Sources` &rarr; `Integrations`
+page (See [Adding a Configuration Field to an Adapter Instance](#adding-a-configuration-field-to-an-adapter-instance)
+and [Adding a Credential](#adding-a-credential). After the account has been created, configuration fields will be
+available in the input to the `collect`, `test`, and `get_endpoint` methods.
+(See [Creating an Adapter Instance](#creating-an-adapter-instance).)
 
-## Adding a Configuration Field to an Adapter Instance
+## Adding a Configuration Field to an Adapter Instance in the Object Model
```
kjrokos, line 56: Any time you change a header like this, we need to make sure all of the anchor links are updated, otherwise they will no longer work. For example, on line 51 we have this line:

`
[Adding a Configuration Field to an Adapter Instance](#adding-a-configuration-field-to-an-adapter-instance)
`

which links to the old version of this header. It will have to be changed to this:

`
[Adding a Configuration Field to an Adapter Instance](#adding-a-configuration-field-to-an-adapter-instance-in-the-object-model)
`

This will have to be done for all of the header updates. Some markdown plugins will detect broken links and help you fix them, I'd recommend making sure you have one that does and the feature is turned on.

doc/adding_to_an_adapter.md
```diff
@@ -1,120 +1,241 @@
...
+Once an adapter instance is defined, any credential defined in the [object model](#defining-an-adapter-and-adapter-instance-in-the-object-model)
+will be prompted to the user when creating an account in VMware Aria Operations on the `Data Sources` &rarr; `Integrations`
+page. After the account has been created, credential fields will be available `AdapterInstance` object passed to the
+`collect`, `test`, and `get_endpoints` methods.(See [Creating an Adapter Instance](#creating-an-adapter-instance).)
+
+Using the [VMware Aria Operations Integration SDK library](https://pypi.org/project/vmware-aria-operations-integration-sdk-lib/),
+the credentials are available in the [AdapterInstance](../lib/python/doc/aria/ops/adapter_instance.html) object passed to
+the `collect`, `test`, and `get_endpoint` methods in `app/adapter.py`(See [Creating an Adapter Instance](#creating-an-adapter-instance).).
+If an adapter supports multiple credentials, **AdapterInstance.get_credential_type** function returns the type of the
+credential used by this adapter instance, or `None` if the adapter instance does not have a credential.
```
kjrokos, line 229:
```suggestion
If an adapter supports multiple credential types, the **AdapterInstance.get_credential_type()** function can be used to determine the type of the
credential used by the adapter instance.
```

doc/references.md
```diff
@@ -23,8 +23,8 @@ Contents
   * describe.xml
   * commands.cfg
 
-* VMware Aria Operations Integration Library
-  * [Python](python-integration-module.md)
+* VMware Aria Operations Integration SDK Library
+  * [Python](https://pypi.org/project/vmware-aria-operations-integration-sdk-lib/)
     * Object model
       * Adapter Instance
         * Identifiers
```
kjrokos, line 30: Can we update the outline to match the current structure of the integration-sdk-lib page? and possibly link within the outline the page for high-level sections if applicable

doc/adding_to_an_adapter.md
```diff
@@ -1,120 +1,241 @@
...
+method will be ignored.
+
+## Defining an Adapter and Adapter Instance in the Object Model
+To define an adapter in the `conf/describe.xml` file use the top-level `AdapterKind` element. Only one adapter can be
+defined in the [object model](#defining-an-adapter-and-adapter-instance-in-the-object-model). The `key` will be used when creating objects (See [Creating an object](#creating-an-object)),
+and must also be present in the `manifest.txt` file in the `"adapter_kinds"` array. When defining an adapter, we also
+have to define an adapter instance. An adapter instance is a special object in VMware Aria Operations that stores user
+configuration for a connection. Every adapter must have exactly one adapter instance type. Adapter instances are set by
+defining a `ResourceKind` element with attribute `type=7`. in Python,
+we can use the [AdapterDefinition](../lib/python/doc/aria/ops/definition/adapter_definition.html).
```
kjrokos, line 28: I think there's still some places we are saying 'adapter instance' where we mean 'adapter instance type'

doc/references.md
kjrokos: I think I was thinking this was something other than what it actually was. I was thinking it was a link to the 'adding to an adapter' page that had been modified, and wanted to make sure that the outline matched the new titles/sections, and links were good.

I think it's good that the outline matches now, but I'm not sure that we should be linking to html pages that don't render properly.

doc/adding_to_an_adapter.md
kjrokos: "An adapter instance type is a special object" should go back to just saying "An adapter instance is a special object" - we need to make sure we're distinguishing between what is an object (resource) in Aria Ops, vs an class/type of object (resourceKind).

The other changes here look correct.

kjrokos (approved):

PR #189, opened 2023-05-24
Python Docs

quirogas:
- Create Mkdocs configuration
- Generate Python documentation using MKdocs
- Update Python documentation docstring style to Google-style docstrings
- Documentation references
- Documentation generation through Guthub actions.


kjrokos:
image

* Page title should have `VMware` before `Aria Operations Integration SDK`
* Same with the left pane title
* Library name is not capitalized correctly - I wonder if we should call it something like `Adapter Library (Python)` instead

PR #203, opened 2023-06-02
Documentation wip

### LINKED ISSUE #187: Github pages site 1.0

quirogas:
https://vmware.github.io/vmware-aria-operations-integration-sdk/

Resolves #187


kjrokos:
Notes: 
 - [ ] Home -> Where Should I Start -> 2nd bullet point: 'Guides' should be a link
 - [ ] Home -> Contributing: There is a link that isn't a link
 - [ ] Guides: Let's reorder the pages so that 'Adding Content' is last and 'Adding to an Adapter' is second to last.
 - [ ] Guides -> Adding Content -> The XML snippets are somewhat hard to read, because you need to scroll horizontally. Is there any option to automatically wrap and keep indents nice?
 - [ ] Guides -> Adding Content -> Adding Localization: At the very end of the file, we have an example of a description key and then show it in the UI. Unfortunately, the original description used `vROps`, and so it got updated at some point without a corresponding update to the screenshot. I think the easiest thing to do is revert the text to match the screenshot.
 - [ ] Guides -> Adding to an Adapter -> What is the Object Model: Can we link the 'Python Adapter Library' to our reference documentation instead of the PyPi Site?
 - [ ] Guides -> Adding to an Adapter -> describe.xml vs AdapterDefinition: Header shouldn't have a question mark, and the first sentence needs a capital letter.
 - [ ] Guides -> Adding to an Adapter: There's a number of places where we use `VMware Aria Operations Integration SDK library`. Let's standardize on 'Python Adapter Library', and link to our documentation site rather than PyPi
 - [ ] Guides -> Creating a New Management Pack/Extending an Existing Management Pack -> Next Steps: Some of these links need to be updated
 - [ ] References -> config.json: This document needs to be cleaned up quite a bit. Let's just remove it for now.

kjrokos (approved):

PR #204, opened 2023-06-05
Documentation/add documentation for connections

### LINKED ISSUE #152: Add documentation for connections
Need to include:
- What is a connection?
- How they are stored
- Editing a connection
- Manually editing a connection

quirogas:
resolves #152

### DIFF
--- a/docs/references/connections.md
+++ b/docs/references/connections.md
@@ -10,15 +10,39 @@ Operations uses this data to help automate and simplify operations management.
 Creating a connection usually involves specifying the IP address or hostname of the external system, along with the appropriate
 [credentials](../guides/adding_to_an_adapter.md#defining-a-credential-in-the-object-model) to access that system. After establishing the connection, an Adapter can collect data from the connected system.
 
+!!! warning
+
+    Credentials for connections are stored in plain text in the 
+    [project config file](project_config.md). For this reason, we recommend that this 
+    file is not included in version control (it is included in the .gitignore file by 
+    defualt), and credentials should be revokable and have the minimum set of
+    permissions necessary for the Adapter to function (Usually Read-Only is sufficient).
+
+In addition, sometimes it is desirable for an Adapter to query VMware Aria Operations' 
+`SuiteAPI`. To facilitate using the API, credentials and connection information are 
+automatically provided to the Adapter when running on a Cloud Proxy. To mimic this when 
+running locally using `mp-test`, `mp-test` needs the SuiteAPI hostname, username, and 
+password.
+
 ??? note
 
-    Suite API connections are handled per project, configuration, unless a connection has its own Suite API credentials. To learn more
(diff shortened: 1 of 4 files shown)


docs/references/connections.md
```diff
@@ -0,0 +1,36 @@
...
+parsing over the object model and using any [TODO]() and [credentials](). The new connection is then stored in the [config.json](config.json) in the root of the project. 
+
+
+#### Editing Existing Connections
+
+Connections can be edited by modifying the key-values in the [config.json](config.json) file at the root of the project that correspond to the value you want to modify.
+
+#### Deleting Connecitons
+
+To delete a connection remove the collection object from [config.json](config.json) file at the root of the project that correspond to the connection you want to delete.
```
kjrokos, line 33: These could probably use a bit more detail. An example could help as well.

docs/references/connections.md
```diff
@@ -0,0 +1,233 @@
...
+                      "part_of_uniqueness": true,
+                      "required": true,
+                      "value": "443"
+                  }
+              },
+              "name": "medium-memory",
+              "suite_api_hostname": null,
+              "suite_api_password": null,
+              "suite_api_username": null
+          },
```
kjrokos, line 94: I think we could make these examples a bit smaller by removing the middle one, and then deleting either the small or large in the next example, without losing any context

docs/references/connections.md
```diff
@@ -0,0 +1,233 @@
...
+
+### What are Connections?
+A **connection** refers to the link between an Adapter and other external systems, such as a vCenter Server, other
+databases, or cloud services. These connections allow an Adapter to gather performance and capacity data about your
+virtual or physical infrastructure operations, which can then be analyzed and presented in a unified, comprehensive view. Aria
+Operations uses this data to help automate and simplify operations management.
+
+Creating a connection usually involves specifying the IP address or hostname of the external system, along with the appropriate
+[credentials]() to access that system. After establishing the connection, an Adapter can collect data from the connected system.
+
```
kjrokos, line 12: We might want to discuss suite-api connection info here as well, since that's a part of a connection. We discuss it a bit in the project config page, but that's more of a 'how are they stored/used', whereas here I think we'd want more of a 'what are they/when do we need them'.

docs/references/project_config.md
```diff
@@ -0,0 +1,219 @@
...
+              }
+            ]
+        ...
+        suite_api_hostname: "my-default-host.example.com",
+        suite_api_password: "my_password",
+        suite_api_username: "my_username",
+     }
+    ```
+
+#### `suite_api_username` (string)
```
kjrokos, line 186: I think this should be at the same level as the password and hostname fields
```suggestion
### `suite_api_username` (string)
```

kjrokos (approved):

PR #205, opened 2023-06-05
How to set up docker hub FAQ (update)

### LINKED ISSUE #199: Documentation: How to set up dockerhub for mp-build (FAQ)
Ensure that we have a documented workflow for setting up a DockerHub registry, as the steps are not intuitive

quirogas:
resolves #199

### DIFF
--- a/docs/troubleshooting_and_faq.md
+++ b/docs/troubleshooting_and_faq.md
@@ -50,9 +50,9 @@ Docker CLI recommends using a token when using docker hub instead of your login 
 3. Run `mp-build`.When prompted about the tag for the container registry, use the following format:
 
     ``` {.shell .copy}
-    docker.io/USER_NAME/CONTAINER_REGISTRY
+    docker.io/USER_NAME/CONTAINER_REPOSITORY
     ```
-The `USER_NAME` should be the same username used to login into docker hub in step 3, and the `CONTAINER_REGISTRY` should be the registry created in step one.
+The `USER_NAME` should be the same username used to login into docker hub in step 3, and the `CONTAINER_REPOSITORY` should be the registry created in step one.
 After entering the tag, you will be prompted to enter your credentials to log into Docker Hub. Enter the same credentials used in step 2.
 
 ???+ note


docs/troubleshooting_and_faq.md
```diff
@@ -36,12 +36,28 @@ running `mp-build`.
 
 Docker CLI recommends using a token when using docker hub instead of your login password, so users should authenticate their Docker Hub account before running `mp-build`.
 
-1. Generate a [dockerhub token](https://hub.docker.com/settings/security?generateToken=true).
-2. Open the `config.json` file in the project's root directory, then replace the key-value of `docker_registry` with the tag of
-   the Docker Hub repository prepended with `docker.io`. For example, if the docker tag is `username/docker-registry-test:tagname`
-   then the key-value will be `docker.io/username/docker-registry-test`.
+1. Create a docker hub container registry with the same name as the docker string in the `adapter_kinds`  key inside the `manifest.txt`  file.
```
kjrokos, line 39: * I don't think we need or want to provide step-by-step instructions for this, but it might be helpful to prefix with something like "Go to hub.docker.com and [...]"
* I don't think the registry has to match the adapter kind name, does it? As long as it's the same name we use in step 3. If that's the case, it might be better to only recommend using the adapter kind name

docs/troubleshooting_and_faq.md
```diff
@@ -36,12 +36,28 @@ running `mp-build`.
...
+The `USER_NAME` should be the same username used to login into docker hub in step 3, and the `ADAPTER_KINDS_KEY` should be the same one used in step 1.
+After entering the tag, you will be prompted to enter your credentials to log into Docker Hub. Enter the same credentials used in step 2.
+
+???+ note
+
+    If `mp-build` doesn't prompt for a tag for the container registry, open the `config.json` file in the project's root directory, then replace the key-value of `docker_registry` with the tag.
+
+!!! warning
+
+    VMware Aria Operations only supports anonymous pulling of images, which may cause issues when using Docker Hub since there is a [Donwload rate limit](https://docs.docker.com/docker-hub/download-rate-limit/#:~:text=Docker%20Hub%20limits%20the%20number,pulls%20per%206%20hour%20period).
```
kjrokos, line 60: Is there another FAQ we could link to about what to do if a user does run into this?

Should we put this at the top so that it's more visible?

docs/troubleshooting_and_faq.md
kjrokos: Sure, but you're very specific about using a particular key in a particular file - in addition, I think there's slightly different requirements for an adapter key vs a registry name. For example, this is a valid adapter key: `MyVeryCoolAdapterKey`, but registries have to be lowercase, and `myverycooladapterkey` is much harder to read - a user might want to change to something like `my_very_cool_adapter_key`, etc

kjrokos (approved):

PR #213, opened 2023-06-08
Feature/support docker hub

### LINKED ISSUE #129: Dockerhub support
Docker cli uses docker hub as its default registry, which causes the tool to use some default logic when outputting login arguments. mp-build requires to use docker.io when using Dockerhub as the container registry. We should find a way to detect when the user is trying to use dockerhub as their default registry instead of them having to input the whole path.

quirogas:
- Allows users to use Docker Hub format to push images
- When a user uses Docker Hub, the image is pushed successfully
- Update the use of the term `docker_registry` to a more appropriate name `container_registry`
- Add test for docker_registry logic

resolves #129 

TODO:
- [x] Handle case where the user inputs `docker.io` in the container registry full path
- [x] Create constants for config variables

### DIFF
--- a/tests/test_mp_build.py
+++ b/tests/test_mp_build.py
@@ -1,4 +1,4 @@
-#  Copyright 2022 VMware, Inc.
+#  Copyright 2023 VMware, Inc.
 #  SPDX-License-Identifier: Apache-2.0
 import pytest
 


vmware_aria_operations_integration_sdk/mp_build.py
```diff
@@ -108,48 +109,69 @@ def build_subdirectories(directory: str) -> None:
...
-def get_docker_registry(
+def get_container_registry(
     adapter_kind_key: str,
     config_file: str,
-    docker_registry_arg: Optional[str],
+    container_registry_arg: Optional[str],
     **kwargs: Any,
 ) -> str:
-    docker_registry = get_config_value("docker_registry", config_file=config_file)
+    container_registry = get_config_value("container_registry", config_file=config_file)
```
kjrokos, line 169: I don't think we should do this without a migration in place, as it will invalidate any existing config files.

Also, we'll need to update anywhere where we reference it in the documentation.

vmware_aria_operations_integration_sdk/mp_build.py
```diff
@@ -108,48 +109,69 @@ def build_subdirectories(directory: str) -> None:
...
+        login(container_registry=container_registry, **kwargs)
+
     except LoginError:
         return False
 
     return True
 
 
+def _is_docker_hub_registry_format(registry: str) -> bool:
+    # if pattern for dockechub is username/repo
```
kjrokos, line 139:
```suggestion
    # if pattern for dockerhub is username/repo
```

vmware_aria_operations_integration_sdk/docker_wrapper.py
```diff
@@ -43,9 +43,16 @@
...
+def login(**kwargs: Any) -> str:
+    command = ["docker", "login"]
+    if "docker_registry" not in kwargs:
+        print(f"Login into Docker Hub")
+        docker_registry = "docker.io"
+    else:
+        docker_registry = kwargs["docker_registry"]
+        print(f"Login into {docker_registry}")
+        command.append(docker_registry)
+
```
kjrokos, line 55: this is called with keyword 'container_registry' now, but we're still checking for 'docker_registry' here.

vmware_aria_operations_integration_sdk/docker_wrapper.py
```diff
@@ -43,9 +43,16 @@
...
-    command = ["docker", "login", f"{docker_registry}"]
+def login(**kwargs: Any) -> str:
+    command = ["docker", "login"]
+    if "docker_registry" not in kwargs:
+        print(f"Login into Docker Hub")
+        docker_registry = "docker.io"
+    else:
+        docker_registry = kwargs["docker_registry"]
+        print(f"Login into {docker_registry}")
+        command.append(docker_registry)
```
kjrokos, line 54: does this append need to happen for both cases?

vmware_aria_operations_integration_sdk/mp_build.py
```diff
@@ -108,48 +109,69 @@ def build_subdirectories(directory: str) -> None:
...
+    pattern = r"^[\w\-]+\/[\w\-]+$"
```
kjrokos, line 140: `_` and `.` are also allowed. Also, should we check that the user hasn't added the full 'docker.io/namespace/repository' path?

vmware_aria_operations_integration_sdk/mp_build.py
kjrokos: I'd rather just write the bit of code that falls back on the old key. Less disruptive and I think less work overall

vmware_aria_operations_integration_sdk/mp_build.py
kjrokos: ah, you're right, it does match `_` -- however, `\w` also matches uppercase letters, which I believe aren't allowed.

kjrokos: I'm not sure what you mean by 'isn't valid outside of the tag'. I can create a repository with `.`s in it:
image

tests/test_mp_build.py
```diff
@@ -0,0 +1,92 @@
+#  Copyright 2022 VMware, Inc.
```
kjrokos, line 1: You should change your copyright header template to the current year!

kjrokos (approved):

Issue #114, opened 2023-04-19
mp-build should prompt for a container registry when the container registry is an empty string in config.json


quirogas:
### Describe the bug

When running mp-build, the container registry field can't be left blank. However, a user can manually edit the `config.json` file and leave an empty string, which causes mp-build to log into the Docker hub and then fail.

![image](https://user-images.githubusercontent.com/22756465/233171686-7ac9afbd-fcf6-43d6-985d-03487ee77264.png)

### Reproduction steps

1. edit the `config.json` file. Enter an empty string for `docker_registry` key value. 
```
{
...
 "docker_registry": "",

}
```
2. run `mp-build`

### Expected behavior

We should prompt the user to enter a registry.

MarkArbogast:
Idea: don't write config.json until we've verified that the registry is valid

PR #214, opened 2023-06-09
container_registry in config file 

quirogas:
- save config value if pushing succeeds 
- We shouldn't log into the Docker Hub when the user provides an empty string because we would still need the names space and repository name.
- related #114

### DIFF
--- a/vmware_aria_operations_integration_sdk/mp_build.py
+++ b/vmware_aria_operations_integration_sdk/mp_build.py
@@ -188,7 +188,6 @@ def validate_container_registry(
     container_registry: Optional[str],
     **kwargs: Any,
 ) -> str:
-
     if not container_registry:
         default_registry_value = get_config_value(
             CONFIG_DEFAULT_CONTAINER_REGISTRY_PATH_KEY


vmware_aria_operations_integration_sdk/mp_build.py
```diff
@@ -184,23 +183,14 @@ def registry_prompt(default: str) -> str:
 def get_container_registry(
     adapter_kind_key: str,
     config_file: str,
+    container_registry: Optional[str],
     container_registry_arg: Optional[str],
```
kjrokos, line 187: Does it make sense to have both of these arguments? If so, we should rename so it's clear how they differ.

Does the function name still make sense? It seems we aren't getting the container registry, except in the case where it doesn't exist yet.

kjrokos: ok, the renamed variables definitely make more sense, but I'm still not convinced we need both. It looks like there's a lot of duplicate code that could be handled by just sending the correct registry value to the function in the first place. There's already fallback logic for the two config keys, just add a new highest-priority case that gets the registry from the command line arguments.

I think if you're going to change the name of the function, it should be in this MR, as this is where the changes that require the new name is happening.

kjrokos (approved):

PR #215, opened 2023-06-09
MP build update

### LINKED ISSUE #114: mp-build should prompt for a container registry when the container registry is an empty string in config.json

### LINKED ISSUE #129: Dockerhub support

### LINKED ISSUE #200: Tag validation

When a user enters an invalid tag format during `mp-build` setup, `mp-build` will try to use the tag, which will cause it to fail. The tool correctly handles the error; however, the message displayed comes from docker and is unclear.

```
17:00:19,405 vmware_aria_operations_integration_sdk.mp_build ERROR Unable to build pak file
17:00:19,407 vmware_aria_operations_integration_sdk.mp_build ERROR ERROR: Unable to build Docker file at /Users/user_name/management-pack:
 {'message': 'invalid reference format'}
```

If we could validate that the tag has a valid format, we could prevent the user from running into this case.

_No response_

### Additional context

_No response_

quirogas:
- Update Documentation FAQ
- Update Documentation for key `docker_registry`
- Update the get_container_registry() function name to reflect its new function
- Removed comment about docker.io port in favor of tracking it as an issue. (https://github.com/vmware/vmware-aria-operations-integration-sdk/issues/220)

Resolves:
- resolves #114
- resolves #200 
- resolves #129

### DIFF
--- a/vmware_aria_operations_integration_sdk/mp_build.py
+++ b/vmware_aria_operations_integration_sdk/mp_build.py
@@ -379,7 +379,6 @@ async def build_pak_file(
                 **kwargs,
             )
 
-            # docker daemon seems to have issues when the port is specified: https://github.com/moby/moby/issues/40619
             conf_registry_field = domain if not port else f"{domain}:{port}"
             conf_repo_field = path
 
--- a/vmware_aria_operations_integration_sdk/validation/input_validators.py
+++ b/vmware_aria_operations_integration_sdk/validation/input_validators.py
@@ -285,8 +285,8 @@ class ContainerRegistryValidator(NotEmptyValidator):
         if match := re.fullmatch(cls.regex, container_registry):
             # the regex group can't distinguis between host and path when no port is specified
             groups = match.groupdict()
-            if not groups["host"]:
-                groups["host"], groups["path"] = (
+            if not groups["domain"]:
+                groups["domain"], groups["path"] = (
                     groups["path"].split("/")[0],
                     groups["path"][1:],
                 )


vmware_aria_operations_integration_sdk/mp_build.py
```diff
@@ -277,30 +370,29 @@ async def build_pak_file(
...
-                build_image(docker_client, path=project_path, tag=f"{registry_tag}")
 
-            with Spinner(f"Pushing Adapter Image to {registry_tag}"):
-                digest = push_image(docker_client, registry_tag)
+                # The first item is the Image object for the image that was built.
+                # The second item is a generator of the build logs as JSON-decoded objects.
+                image, _ = build_image(docker_client, path=project_path)
+
+            # We have to add the Optional type anotation otherwise mypy will mark any line after the while loop as unreachable
+            # https://github.com/python/mypy/issues/5423
```
kjrokos, line 380: I don't see a while loop around here - did the code get moved without its comment?

vmware_aria_operations_integration_sdk/mp_build.py
```diff
@@ -277,30 +370,29 @@ async def build_pak_file(
...
+
+            conf_repo_field, conf_registry_field, digest = _tag_and_push(
```
kjrokos, line 382: Are the order of these correct? `tag_and_push` returns `domain, path, digest`, which makes me think registry and repo should be switched.

vmware_aria_operations_integration_sdk/mp_build.py
```diff
@@ -108,48 +124,138 @@ def build_subdirectories(directory: str) -> None:
...
+    if not container_registry:
+        container_registry = get_config_value(
+            CONFIG_FALLBACK_CONTAINER_REGISTRY_KEY, config_file=config_file
+        )
+
+    # we want to keep track of the original value, so we can update it if nesessary
+    original_value = container_registry
+
+    digest = ""
+    should_prompt = container_registry is None
```
kjrokos, line 187: IMO, we should prompt anytime the push fails because of an invalid tag, _except_ if the tag was specified via the command line argument.

vmware_aria_operations_integration_sdk/mp_build.py
```diff
@@ -108,48 +124,138 @@ def build_subdirectories(directory: str) -> None:
...
+            **kwargs,
+        )
+        tag = manifest["version"] + "_" + str(time.time())
+
+        # docker daemon seems to have issues when the port is specified: https://github.com/moby/moby/issues/40619
+
+        registry_tag = f"{container_registry}:{tag}"
+        logger.debug(f"registry tag: {registry_tag}")
+
+        image.tag(container_registry, tag)
```
kjrokos, line 203: Can we make it more clear what the two tags (`registry_tag` and `tag`) are each for, and why we're building the `registry_tag` but then only tagging `tag` here?

I think a comment is probably sufficient

vmware_aria_operations_integration_sdk/mp_build.py
kjrokos: You used `REPOSITORY` for both parts above... Regardless, I think it would clear things up if we used the same terminology in both places.

PR #229, opened 2023-07-07
Feature/mp init blank template

### LINKED ISSUE #121: Option in mp-init for a blank template

When creating a new project, the template adapter sometimes gets in the way, since there is a lot of code that is not relevant.

When creating a new project, we should have a two options:
* Template adapter
* Blank adapter

The blank adapter should have stubs for all methods, including generic try/catch/finally and return statements, but no code that needs to be replaced. In addition, we should make sure the `adapter_requirements.txt` file removes psutil for the blank adapter.

The three options are:
* Keep as-is: Only have the template adapter. Downside is manual work whenever creating an new adapter
* Switch to only the blank adapter. Downside is there is no reference when first starting with the SDK. Some people may prefer to modify code that works rather than starting from blank methods
* Have both, and allow the user to select when running `mp-init`. This is my preferred solution.

### Additional context

_No response_

quirogas:
- Add a blank template adapter project option for mp-init
- Update documentation

image

resolves #121

### DIFF
--- a/docs/get_started.md
+++ b/docs/get_started.md
@@ -105,7 +105,7 @@ the creation of a new management pack project.
      `New Adapter`: Creates methods, minimal comments, and code necessary to implement test
       connection, collection, adapter definition, and endpoints logic.
 
-     If you are following along the *Get Started*, select **Sample Adapter**.
+     For the purposes of this *Get Started* guide, select **Sample Adapter**.
 
 ???+ info
 


vmware_aria_operations_integration_sdk/mp_init.py
```diff
@@ -465,9 +494,16 @@ def build_project_structure(
...
             )
 
-        # copy adapter.py into app directory
-        with resources.path(adapter_template, "adapter.py") as src:
-            dest = os.path.join(path, project_directory)
+        # copy the template code into app/adapter.py file
+        template = (
+            "adapter.py"
+            if template_style == TEMPLATE_ADAPTER_ENTRYPOINT_OPTION_KEY
+            else "blank_template_adapter"
```
kjrokos, line 501: Does this work? It looks like it needs a '.py' at the end?

vmware_aria_operations_integration_sdk/adapter_template/blank_template_adapter.py
```diff
@@ -0,0 +1,169 @@
...
+    validate, process, and display the data correctly.
+    :return: AdapterDefinition
+    """
+    with Timer(logger, "Get Adapter Definition"):
+        definition = AdapterDefinition(ADAPTER_KIND, ADAPTER_NAME)
+
+        # The key 'container_memory_limit' is a special key read by the VMware Aria Operations
+        # collector to determine how much memory to allocate to the docker container running this
+        # adapter. It does not need to be read inside the adapter code. If the adapter doesn't
+        # read the key, the container running this adapter will be given 1024 MB of memory.
```
kjrokos, line 32: The adapter never needs to read the key. It's Aria Ops that reads the key. (By the time the adapter is running, the container is already created, since the adapter is running inside it)

vmware_aria_operations_integration_sdk/adapter_template/blank_template_adapter.py
```diff
@@ -0,0 +1,169 @@
...
+            "container_memory_limit",
+            label="Adapter Memory Limit (MB)",
+            description="Sets the maximum amount of memory VMware Aria Operations can "
+            "allocate to the container running this adapter instance.",
+            required=True,
+            advanced=True,
+            default=1024,
+        )
+
+        # TODO: Add definitions
```
kjrokos, line 43: I think we probably want some less generic TODOs in this function.

vmware_aria_operations_integration_sdk/adapter_template/blank_template_adapter.py
```diff
@@ -0,0 +1,169 @@
+#  Copyright 2022 VMware, Inc.
```
kjrokos, line 1: 2023

docs/get_started.md
```diff
@@ -96,16 +96,20 @@ the creation of a new management pack project.
 
 For complete documentation of the `mp-init` tool including an overview of its output, see the [MP Initialization Tool Documentation](references/mp-init.md).
 
-### Template Project
+### Template Project and Entry Point
```
kjrokos, line 99:
```suggestion
### Template Projects
```

vmware_aria_operations_integration_sdk/adapter_template/blank_template_adapter.py
```diff
@@ -0,0 +1,169 @@
...
+        try:
+            # A typical test connection will generally consist of:
+            # 1. Read identifier values from adapter_instance that are required to
+            #    connect to the target(s)
+            # 2. Connect to the target(s), and retrieve some sample data
+            # 3. Disconnect cleanly from the target (ensure this happens even if an
+            #    error occurs)
+            # 4. If any of the above failed, return an error, otherwise pass.
+
+            # TODO: Add collection testing logic
```
kjrokos, line 61:
```suggestion
            # TODO: Add connection testing logic
```

docs/get_started.md
```diff
@@ -96,16 +96,20 @@ the creation of a new management pack project.
...
-`adapter.py` is a template adapter that collects several objects and metrics from the
-container in which the adapter is running. The template adapter has comments throughout its code that explain what the code does
-and how to customize it for your adapter.
+- `Boilerplate Entrypoint`: Ideal for experienced users;
+  it comes without any examples, just the necessary code and comments to implement test connection,
+  collection, adapter definition, and endpoints logic.
 
-The methods inside the adapter template are required, and should be modified to generate a custom
+The methods inside the `adapter.py` are required, and should be modified to generate a custom
 adapter. Each method fulfills a request from the VMware Aria Operations collector, and can be tested individually using
```
kjrokos, line 113: * The entry point isn't changing - it's `adapter.py` in both cases, so I think most of the references to entry point are not needed.
* I think both options could reasonably be called 'templates', so let's not use that for distinguishing between the two.

Given the above, how about something like `Sample Adapter` or `Example Adapter` for the psutil adapter. For the 'blank' adapter here are some suggestions - we can talk about pros and cons offline with Mark:
* `Blank Adapter`
* `Default Adapter`
* `Stub Adapter`
* `New Adapter`
*  for the other. I don't think [Boilerplate](https://en.wikipedia.org/wiki/Boilerplate_code#:~:text=In%20computer%20programming%2C%20boilerplate%20code,to%20accomplish%20only%20minor%20functionality.) fits here.

vmware_aria_operations_integration_sdk/mp_init.py
```diff
@@ -319,8 +326,28 @@ def main() -> None:
...
+                    "Boilerplate Entrypoint (Advanced users)",
+                ),
+            ],
+            description="- `Template Adapter Entry Point`: Ideal for first-time users; it comes with a template adapter\n"
+            "that collects several objects and metrics from the container the adapter is running. The template\n"
+            "adapter has comments throughout its code that explain what the code does and how to customize\n"
+            "it for your adapter.\n "
+            "- `Boilerplate Entrypoint`: Ideal for experienced users; it comes with the necessary code\n"
+            "and comments to implement test connection, collection, adapter definition, and endpoints logic.\n\n"
+            "For additional information visit https://vmware.github.io/vmware-aria-operations-integration-sdk/get_started/#template-project-and-entry-point",
```
kjrokos, line 347: I think this is a bit wordy for help text, especially since we're linking to documentation that essentially has the same wording.

docs/get_started.md
kjrokos: (Make sure links to this get changed as well)

vmware_aria_operations_integration_sdk/adapter_template/blank_template_adapter.py
kjrokos: Yes, that is correct

docs/get_started.md
MarkArbogast: I prefer `New Adapter`

docs/get_started.md
```diff
@@ -97,16 +97,20 @@ the creation of a new management pack project.
...
-container in which the adapter is running. The template adapter has comments throughout its code that explain what the code does
-and how to customize it for your adapter.
+- `New Adapter`: Ideal for experienced users;
+  it comes without any examples, just the necessary code and comments to implement test connection,
+  collection, adapter definition, and endpoints logic.
 
-The methods inside the adapter template are required, and should be modified to generate a custom
+The methods inside the `adapter.py` are required, and should be modified to generate a custom
 adapter. Each method fulfills a request from the VMware Aria Operations collector, and can be tested individually using
 `mp-test` (covered in [Testing a Management Pack](#testing-a-management-pack)).
```
kjrokos, line 115: I think we need to modify the structure of this a bit. Previously, the discussion about the template adapter applied to every mp-init output, so it was after the walkthrough. However, we should move the selection part up to a 'step 8' in the "Creating a Management Pack" section and discuss the options there, then move into some of the details of how projects and adapters are setup and work (in a generic way) here.

I'm also unsure if we should market the 'New Adapter' as 'Ideal for experienced users' and the sample adapter as 'Ideal for first time users'. I think it's better to just describe what they are, but in the walkthrough section above make sure we direct the user to select 'Sample Adapter'.

vmware_aria_operations_integration_sdk/mp_init.py
```diff
@@ -433,7 +457,8 @@ def build_project_structure(
         # create template requirements.txt
         requirements_file = os.path.join(path, "adapter_requirements.txt")
         with open(requirements_file, "w") as requirements:
-            requirements.write("psutil==5.9.4\n")
+            if template_style != NEW_ADAPTER_OPTION_KEY:
+                requirements.write("psutil==5.9.4\n")
```
kjrokos, line 461: I would invert this to test for equality to the sample adapter -- if we ever have additional templates/options, the sample adapter is likely the only one that will require psutil

vmware_aria_operations_integration_sdk/mp_init.py
```diff
@@ -465,9 +490,20 @@ def build_project_structure(
...
-            dest = os.path.join(path, project_directory)
+        namespace_package_indicator_file = os.path.join(path, "__init__.py")
+        with open(namespace_package_indicator_file):
+            os.utime(namespace_package_indicator_file)
+
+        # copy the template code into app/adapter.py file
+        template = (
+            "adapter.py"
+            if template_style == SAMPLE_ADAPTER_OPTION_KEY
+            else "blank_template_adapter.py"
```
kjrokos, line 501: We may want to rename the files to match the descriptions we decided on.

docs/get_started.md
MarkArbogast: I also wouldn't say `necessary code`. Open to ideas here, but maybe something more along the lines of `minimum code` or `minimal code`.

docs/get_started.md
```diff
@@ -95,18 +95,26 @@ the creation of a new management pack project.
...
+      The template adapter has comments throughout its code that explain what the code does and how to customize it for your adapter.
+
+     `New Adapter`: Creates methods, minimal comments, and code necessary to implement test
+      connection, collection, adapter definition, and endpoints logic.
 
-`adapter.py` is a template adapter that collects several objects and metrics from the
-container in which the adapter is running. The template adapter has comments throughout its code that explain what the code does
-and how to customize it for your adapter.
+     Select **Sample Adapter** if you are following along the walkthrough.
+     For complete documentation of the `mp-init` tool including an overview of its output,
```
kjrokos, line 109: This should be in a separate paragraph outside of the numbered steps

docs/get_started.md
```diff
@@ -95,18 +95,26 @@ the creation of a new management pack project.
...
+     see the [MP Initialization Tool Documentation](references/mp-init.md).
+
+### Project Templates
+Both templates create a file system that has the basic project structure required to develop and build a Management Pack.
+Each file and directory is discussed in depth in the [mp-init](references/mp-init.md) documentation. `app/adapter.py` is the adapter's
+entry point and the best starting point. Users can pick between two entry points when creating their project:
```
kjrokos, line 115:
```suggestion
entry point and the best starting point. 
```

docs/get_started.md
```diff
@@ -95,18 +95,26 @@ the creation of a new management pack project.
...
 
-The methods inside the adapter template are required, and should be modified to generate a custom
+The methods inside the `adapter.py` are required, and should be modified to generate a custom
```
MarkArbogast, line 117: `the adapter.py` -> `adapter.py`

MarkArbogast (approved):
Approved pending removal of extra `the`

kjrokos (approved):

PR #243, opened 2023-07-20
Split config files

### LINKED ISSUE #130: Split config.json into shareable and non-shareable
There is some info we want from config file (e.g., container registry) that would be nice to share

quirogas:
- Split config.json file into -> `config.json` and `connections.json`
- Add migration logic so existing users don't have to move connection info manually (the only manual step is to remove config.json from the .gitignore file)
- use constants for all `config.json` and `connection.json` files keys
- Update project_info.md page to reflect new changes
- Add a `Project Connections File` page to the reference section 
- Add an FAQ to the Other page explaining to the user why they are seeing the migration warning message

resolves #130 

## Example:
image

#### config.json
image

#### config.json
image
#### connections.json
image


vmware_aria_operations_integration_sdk/project.py
```diff
@@ -107,27 +109,61 @@ def name(self) -> str:
...
-        local_config_file = os.path.join(path, "config.json")
+        local_config_file = os.path.join(path, CONFIG_FILE_NAME)
+        connections_file = os.path.join(path, CONNECTIONS_FILE_NAME)
+
+        if not os.path.isfile(connections_file):
+            if os.path.isfile(
+                local_config_file
+            ):  # if there is a config.json, but not a connections.json file, is a sign that is an old version of the config.json file
+                logger.warning(
+                    f"Possible {CONFIG_FILE_NAME} file detected. To learn more see https://vmware.github.io/vmware-aria-operations-integration-sdk/troubleshooting_and_faq/collections/#how-can-i-migrate-my-old-configjson-file-deprecated-fields-detected-in-configjson-file-warning"
```
kjrokos, line 131: I believe this will display as:
"Possible config.json file detected. [...]" Which I don't think describes the situation.

Also, not having a "connections.json" file doesn't strike me as necessarily something we should warn about. For example, someone has cloned a repo someone else made and is running it for the first time. They won't have a 'connections.json' file, because we're excluding it from source control, but that doesn't mean there's a problem, just that they haven't created one yet.

I think the warning only makes sense if there's also a "connections" key inside the 'config.json' file. (and even then, it may be more appropriate as just an info log)

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -107,27 +109,61 @@ def name(self) -> str:
...
+                )
+            with open(connections_file, "w") as _connections:
+                json.dump({}, _connections, indent=4, sort_keys=True)
```
kjrokos, line 134: I'd recommend populating with the "connections" data from 'config.json' if it exists.

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -107,27 +109,61 @@ def name(self) -> str:
...
+
+            if (
+                "suite_api_hostname" in json_config
+                or "suite_api_username" in json_config
+                or "suite_api_password" in json_config
+                or "suite_api_connection" in json_config
+                or "connections" in json_config
+            ):
+                logger.warning(
+                    f"Deprecated fields detected in {CONFIG_FILE_NAME}. To learn more see https://vmware.github.io/vmware-aria-operations-integration-sdk/troubleshooting_and_faq/collections/#how-can-i-migrate-my-old-configjson-file-deprecated-fields-detected-in-configjson-file-warning"
```
kjrokos, line 156: I'm confused why this is a separate case from above.

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -199,3 +235,20 @@ def _add_and_update_project_paths(path: str) -> None:
             )
         ),
     )
+
+
+def _safe_append_to_gitignore(gitignore_file_path: str, token: str) -> None:
+    try:
+        with open(gitignore_file_path, "r") as gititnore:
```
kjrokos, line 242: typo: gititnore -> gitignore

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -199,3 +235,20 @@ def _add_and_update_project_paths(path: str) -> None:
...
+            for line in gititnore.readlines():
+                if token in line:
```
kjrokos, line 244: We probably want to match the whole line, rather than just (potentially) a part of it.

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -199,3 +235,20 @@ def _add_and_update_project_paths(path: str) -> None:
...
+                    return
+
+        with open(gitignore_file_path, "a") as gitignore:
+            gitignore.write(f"{token}\n")
+
+        logger.info("Appended {token} to .gitignore")
```
kjrokos, line 250:
```suggestion
        logger.info(f"Appended {token} to .gitignore")
```

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -199,3 +235,20 @@ def _add_and_update_project_paths(path: str) -> None:
...
+    except FileNotFoundError:
+        logger.warning(
+            "file not found in {file_path} when attempting to safe append {token}"
```
kjrokos, line 253:
```suggestion
            f"File not found in {file_path} when attempting to append {token}"
```

vmware_aria_operations_integration_sdk/constant.py
```diff
@@ -29,17 +29,29 @@
...
 
-GLOBAL_CONFIG_FILE = path.join(CONFIG_DIRECTORY, "config.json")
-CONFIG_CONTAINER_REGISTRY_KEY = "container_registry"
-CONFIG_FALLBACK_CONTAINER_REGISTRY_KEY = "docker_registry"
+# Configuration files
+CONFIG_FILE_NAME = "config.json"
+GLOBAL_CONFIG_FILE = path.join(CONFIG_DIRECTORY, CONFIG_FILE_NAME)
+CONNECTIONS_FILE_NAME = "connections.json"
+CONNECTIONS_CONFIG_CONTAINER_REPOSITORY_KEY = "container_repository"
+CONNECTIONS_CONFIG_FALLBACK_CONTAINER_REGISTRY_KEY = "docker_registry"
```
kjrokos, line 37: Remove reference to 'Connections' for these

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -107,27 +139,76 @@ def name(self) -> str:
...
+            docker_port = json_config.get(CONFIG_DOCKER_PORT_KEY, 8080)
+
+            for key in connection_file_keys:
+                if key in json_config:
+                    logger.warning(
+                        f"{CONNECTIONS_FILE_NAME} key '{key}' found in {CONFIG_FILE_NAME}."
+                    )
+                    connections_data[key] = json_config.get(key)
+                    logger.debug(f"Deleting {key} from {CONNECTIONS_FILE_NAME}")
+                    del json_config[key]
```
kjrokos, line 184: We don't want to be deleting/transferring things from config files automatically. There should be a confirmation step that says exactly what it's going to do with an option to decline. I'd also not emit multiple warnings - I'd do something like this:
1. Determine if any keys are present in config.json that we want to migrate
2. If there are, prompt user to migrate all of them.

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -107,27 +139,76 @@ def name(self) -> str:
...
+
+            if len(connections_data):
+                _config.seek(0)
+                logger.info(
+                    f"deleting {CONNECTIONS_FILE_NAME} keys from {CONFIG_FILE_NAME}"
```
kjrokos: This seems like the same information in the log line on 183

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -107,27 +139,76 @@ def name(self) -> str:
...
+                )
+                json.dump(json_config, _config, indent=4, sort_keys=True)
+                _config.truncate()
+                logger.info(
+                    f"To learn more about '{CONNECTIONS_FILE_NAME}' file visit TODO: ADD LINK HERE"
```
kjrokos: Put a message like this in the prompt, so users can find more info before they commit to migration.

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -107,27 +139,76 @@ def name(self) -> str:
...
+            CONNECTIONS_CONFIG_SUITE_API_PASSWORD_KEY,
+            CONNECTIONS_CONFIG_SUITE_API_CONNECTION_KEY,
+            CONNECTIONS_CONFIG_CONNECTIONS_LIST_KEY,
+        ]
+
+        with open(local_config_file, "r+") as _config:
+            json_config = json.load(_config)
+            docker_port = json_config.get(CONFIG_DOCKER_PORT_KEY, 8080)
+
+            for key in connection_file_keys:
```
kjrokos, line 177: Let's extract all of the migration code into a separate method, to keep the extract method cleaner

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -199,3 +280,20 @@ def _add_and_update_project_paths(path: str) -> None:
             )
         ),
     )
+
+
+def _safe_append_to_gitignore(gitignore_file_path: str, token: str) -> None:
+    try:
+        with open(gitignore_file_path, "r") as gitignore:
+            for line in gitignore.readlines():
+                if token == line.rstrip("\n"):
```
kjrokos, line 289: would `strip()` be more appropriate here? I'd imagine ignoring any whitespace would be ideal.

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -199,3 +280,20 @@ def _add_and_update_project_paths(path: str) -> None:
...
+                    return
+
+        with open(gitignore_file_path, "a") as gitignore:
+            gitignore.write(f"{token}\n")
+
+        logger.info(f"Appended '{token}' to .gitignore")
+    except FileNotFoundError:
+        logger.warning(
+            f"File not found in {gitignore_file_path} when attempting to append {token}"
+        )
```
kjrokos, line 299: One reason I can think of why we might get here, is if the user doesn't want to use git. For example, if their organization uses Perforce and they want to use that. Given that, we might want to massage this warning message a bit. I'm thinking something like 
"Could not automatically set the file '{token}' to be ignored in version control."

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -78,18 +108,64 @@ def get_memory_limit(self) -> int:
...
+                description=f"All elements related to {CONFIG_FILE_NAME} will be deleted.\n"
+                "If 'Yes' is selected, then the deleted elements will be migrated into"
+                f" {CONNECTIONS_FILE_NAME}.\n"
+                f"If 'No' is selected, then all deleted elements will not be migrated into "
+                f"{CONNECTIONS_FILE_NAME}.\n"
+                "To learn more about connection config file migration, visit\n"
+                f"https://vmware.github.io/vmware-aria-operations-integration-sdk"
+                f"/troubleshooting_and_faq/other"
+                f"/#migrating-connection-related-elements-from-configjson-to-connectionsjson",
+            )
```
kjrokos, line 158: No should be a no-op, not delete without migration

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -78,18 +108,64 @@ def get_memory_limit(self) -> int:
...
+def _read_with_merge_prompt(local_config_file: str) -> tuple[dict[Any, Any], Any]:
+    connections_data = {}
+    connection_file_element_keys = [
+        CONNECTIONS_CONFIG_SUITE_API_HOSTNAME_KEY,
+        CONNECTIONS_CONFIG_SUITE_API_USERNAME_KEY,
+        CONNECTIONS_CONFIG_SUITE_API_PASSWORD_KEY,
+        CONNECTIONS_CONFIG_SUITE_API_CONNECTION_KEY,
+        CONNECTIONS_CONFIG_CONNECTIONS_LIST_KEY,
+    ]
+
```
kjrokos, line 134: Test to see if 'connections.json' exists. If it doesn't, AND config.json has deprecated keys, prompt to migrate. If it does, then exit this function without doing anything

vmware_aria_operations_integration_sdk/project.py
kjrokos: But always create connections.json

kjrokos: Also, this is where we should add the connections.json to .gitignore

vmware_aria_operations_integration_sdk/project.py
```diff
@@ -107,27 +183,44 @@ def name(self) -> str:
...
+        connections_data, docker_port = _read_with_merge_prompt(local_config_file)
+
+        if not os.path.isfile(connections_file) or len(connections_data):
+            with open(connections_file, "w") as _connections:
+                json.dump(connections_data, _connections, indent=4, sort_keys=True)
+
+        _safe_append_to_gitignore(
+            os.path.join(path, ".gitignore"), CONNECTIONS_FILE_NAME
+        )
```
kjrokos, line 212: Move to `_read_with_merge_prompt`

docs/troubleshooting_and_faq/other.md
```diff
@@ -21,3 +21,24 @@ If the pak file installs successfully but errors when creating an account (adapt
 - Check that the Cloud Proxy supports containerized adapters. Containerized adapter
   support is supported in VMware Aria Operations version 8.10.0 and later.
 
+### Migrating connection-related elements from config.json to connections.json?
```
kjrokos, line 24:
```suggestion
### How do I migrate connection-related elements from config.json to connections.json?
```

docs/troubleshooting_and_faq/other.md
```diff
@@ -21,3 +21,24 @@ If the pak file installs successfully but errors when creating an account (adapt
...
+[project connections JSON file](../references/project_connections_config.md)(`connections.json`).
+As part of this change, both `mp-test` and `mp-build` will offer to migrate connection-related
+elements to the `connections.json` file when present in the `config.json` file
+(The new `connections.json` file is also added to the project's `.gitignore`
+to prevent sensitive information from being committed).
+Regardless of whether the user decides to migrate the values or not, the found elements will be deleted from the
+`configuration.json` file.
+Moving all the connection-related information away from the `config.json` file allows users
+to include their project configuration file in version control,
+making using the same `container_repository` for the project easier.
```
kjrokos, line 38:
```suggestion
all connection-related elements from the [project config file](../references/project_config.md)
have been migrated to a new
[project connections JSON file](../references/project_connections_config.md)(`connections.json`).
As part of this change, both `mp-test` and `mp-build` will offer to migrate connection-related
elements to the `connections.json` file when present in the `config.json` file
(The new `connections.json` file is automatically added to the project's `.gitignore`
to prevent sensitive information from being committed).
Moving all the connection-related information away from the `config.json` file allows users
to include their project configuration file in version control,
making using the same `container_repository` for the project easier.
```

kjrokos (approved):

PR #268, opened 2023-07-28
Bug/cli arguments range

### LINKED ISSUE #246: CLI arguments range error message is too long
### Describe the bug

Arguments with ranges display all possible ranges when the choice is outside of the range.

### Reproduction steps

mp-test --port -1

### Expected behavior

usage: mp-test [-h] [-V] [-p PATH] [-c CONNECTION] [-v [0-3]] [-P [0, 65536]]
               {connect,collect,long-run,endpoint_urls,adapter_definition,version,wait} ...
mp-test: error: argument -P/--port: invalid choice: -1 (choose from 0, 65536)

### Additional context

_No response_

quirogas:
- Add better error logging for parameters with possible range

Resolves #246

# Before:
image
# After:
image

### DIFF
--- a/vmware_aria_operations_integration_sdk/util.py
+++ b/vmware_aria_operations_integration_sdk/util.py
@@ -24,6 +24,7 @@ class LazyAttribute(object):
 class RangeAction(argparse.Action):
     def __init__(
         self,
+        *args: Any,
         lower_bound: int = 0,
         upper_bound: int = 100,
         **kwargs: Any,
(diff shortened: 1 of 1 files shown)


vmware_aria_operations_integration_sdk/util.py
```diff
@@ -20,25 +21,53 @@ def __get__(self, instance: Optional[object], owner: Optional[type[object]]) ->
...
+        nargs: Optional[int] = None,
+        const: Any = None,
+        default: Any = None,
+        type: Any = None,
+        choices: Any = None,
+        required: bool = False,
+        help: Optional[str] = None,
+        metavar: Union[str, None] = None,
+        lower_bound: int = 0,
+        upper_bound: int = 100,
```
kjrokos, line 38: * For the arguments we don't specifically use, can we just use `**kwargs`?
* Does it work to set `metavar` here programmatically based on lower_bound and `upper_bound`?
* If so, and we keep `metavar` in the argument list, can it's type be `Optional[str]`?

kjrokos (approved):

PR #272, opened 2023-08-16
Add 'Key' and 'Identifier' classes to Java Adapter Library

kjrokos:
Adds 'Key' class that enforces uniqueness among Objects.
Adds 'Identifier' class, which is a dependency of the 'Key' class
Adds test suite which tests 'Key' and implicitly tests 'Identifier'


lib/java/AdapterLibrary/src/main/kotlin/com/vmware/aria/operations/Key.kt
```diff
@@ -0,0 +1,129 @@
...
+                key.addAll(uniqueIdentifiers.sortedBy { identifier -> identifier.key })
+            }
+            return key
+        }
+
+    /**
+     * @return True if the keys will resolve to the same object, false otherwise.
+     */
+    override fun equals(other: Any?): Boolean {
+        return if (other is Key) {
```
MarkArbogast, line 84: could be simplified to 

```
return other is Key && internalKey == other.internalKey
```

lib/java/AdapterLibrary/src/main/kotlin/com/vmware/aria/operations/Key.kt
```diff
@@ -0,0 +1,129 @@
...
+     * @param defaultValue An optional default value.
+     * @return The value associated with the identifier.
+     * If the value associated with the identifier is empty and 'defaultValue' is
+     * provided, returns 'defaultValue'.
+     * If the identifier does not exist, returns defaultValue if provided, else 'null'.
+    */
+    @JvmOverloads
+    fun getIdentifier(key: String, defaultValue: String? = null): String? {
+        val value = identifiers[key]?.value
+        return if (value?.isBlank() != false) {
```
MarkArbogast, line 109: could simplified to 

```
return if (value?.isBlank()) {
```

MarkArbogast (approved):
Approved pending consideration of a couple simplifications

lib/java/AdapterLibrary/src/main/kotlin/com/vmware/aria/operations/Key.kt
kjrokos: Actually, this is not exactly equivalent, because `value` is nullable. If it is, then `value?.isBlank()` evaluates to `null`, which Kotlin (and Java) will not coerce to a boolean.

PR #279, opened 2023-08-25
Adds Adapter Logger setup functionality

kjrokos:
* Adds `AdapterLogger` class that simplifies setting up and creating a logger in Java adapter code
* Reads a simple configuration from the same `loglevels.cfg` file as the server and Python adapters.

### DIFF
--- a/lib/java/AdapterLibrary/src/main/kotlin/com/vmware/aria/operations/AdapterLogger.kt
+++ b/lib/java/AdapterLibrary/src/main/kotlin/com/vmware/aria/operations/AdapterLogger.kt
@@ -61,7 +61,18 @@ internal fun setupLogging(
     filePath: String,
 ) {
 
-    val pattern = "%d [%t] %-5level: %msg%n%throwable"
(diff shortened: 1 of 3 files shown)


lib/java/AdapterLibrary/src/main/kotlin/com/vmware/aria/operations/AdapterLogger.kt
```diff
@@ -0,0 +1,257 @@
...
+ *  @param filePath The location where the config file and logs reside in.
+ */
+internal fun setupLogging(
+    filename: String,
+    backupFileCount: Int,
+    maxSize: Long,
+    filePath: String,
+) {
+
+    val pattern = "%d [%t] %-5level: %msg%n%throwable"
```
MarkArbogast, line 64: Would be useful to add a comment here with an example log that follows this pattern.

MarkArbogast (approved):
Approved. Consider adding an example log to show the pattern.

PR #280, opened 2023-08-28
Adds Java Library Named Pipe IO methods

kjrokos:
* Like it says on the tin. There's only a single test, because each is dependent on the other (or a similar piece of code, as will be the case when communicating to the server), so it makes sense to test the read and write methods at the same time.


MarkArbogast (approved):
Looks good!

PR #284, opened 2023-09-05
Feature/java default project structure

### LINKED ISSUE #250: Create a directory structure for default adapter templates that includes language
E.g., adapter_template/[language]/[template_type]/

quirogas:
- Add Java project structure
- Add build.gradle file

Closes #250

### DIFF
--- a/vmware_aria_operations_integration_sdk/adapter_config.py
+++ b/vmware_aria_operations_integration_sdk/adapter_config.py
@@ -1,0 +1,26 @@
+from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import List
+
+
+class Question:
+    def __init__(self, key: str, prompt: Callable, *args: Any, **kwargs: Any) -> None:
+        self.key = key
+        self.prompt = prompt
+        self.prompt_args = args
+        self.prompt_kwargs = kwargs
+
+    def ask(self) -> Any:
+        return self.prompt(*self.prompt_args, **self.prompt_kwargs)
+
+
+class AdapterConfig:
+    def __init__(self, language: str, questions: List[Question] = []) -> None:
+        self.language = language
+        self.values: Dict[str, Any] = dict()
+        self.questions = questions
+
+    def prompt_config_values(self) -> None:
+        for question in self.questions:
+            self.values[question.key] = question.ask()
--- a/vmware_aria_operations_integration_sdk/mp_init.py
+++ b/vmware_aria_operations_integration_sdk/mp_init.py
@@ -16,6 +16,8 @@ import pkg_resources
 from git import Repo
 
 from vmware_aria_operations_integration_sdk import adapter_template
+from vmware_aria_operations_integration_sdk.adapter_config import AdapterConfig
+from vmware_aria_operations_integration_sdk.adapter_config import Question
 from vmware_aria_operations_integration_sdk.adapter_template import java
 from vmware_aria_operations_integration_sdk.adapter_template import powershell
 from vmware_aria_operations_integration_sdk.constant import CONNECTIONS_FILE_NAME
@@ -174,9 +176,8 @@ def create_project(
     vendor: str,
     eula_file: str,
     icon_file: str,
-    language: str,
+    adapter_config: AdapterConfig,
     template_style: str,
-    package_name: str,
 ) -> None:
     mkdir(path)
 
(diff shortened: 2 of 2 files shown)


vmware_aria_operations_integration_sdk/mp_init.py
```diff
@@ -345,6 +352,19 @@ def main() -> None:
...
+        if language == "java":
+            package_name = prompt(
+                "Enter package name: ",
+                default="com.mycompany",
+                validator=ChainValidator(
+                    [
+                        NotEmptyValidator("Java Package"),
+                        NoUppercaseValidator("Java Package"),
+                    ]
+                ),
```
kjrokos, line 364: There's nothing wrong with this, but what we've done in the past for this case is to have validators inherit from `NotEmptyValidator` - for example, `NewProjectDirectoryValidator` or `UniquenessValidator`

vmware_aria_operations_integration_sdk/adapter_template/java.py
```diff
@@ -3,28 +3,111 @@
 import os
```
kjrokos, line 3: When we make the final templates, I think we should follow the paradigm we did for the python adapters, where we have a full source file (or files) that get copied over. If we need to add anything dynamically (for example, we will want to specify the package), we can look into python templating engines.

vmware_aria_operations_integration_sdk/mp_init.py
```diff
@@ -567,18 +593,43 @@ def build_project_structure(
...
 
-        mkdir(path, "src")
-        java.build_template(path, "src")
+        project_directory = "src"
+        package_path = os.path.join(project_directory, *package_name.split("."))
+        mkdir(path, package_path)
+        java.build_template(path, package_path)
+
+        # Add build.gradle file
+        with open(os.path.join(path, "build.gradle"), "w") as gradle_file:
```
kjrokos, line 602: The gradle files will almost certainly be different between the different adapter templates. Let's look at moving this to its own file as well when we make the final adapter templates.

vmware_aria_operations_integration_sdk/validation/input_validators.py
```diff
@@ -199,6 +199,16 @@ def validate(self, document: Document) -> None:
             validator.validate(document)
 
 
+class NoUppercaseValidator(Validator):  # type: ignore
```
kjrokos, line 202: Can you add tests for this in `tests/validation/test_input_validators.py`

vmware_aria_operations_integration_sdk/mp_init.py
```diff
@@ -357,6 +377,7 @@ def main() -> None:
                 icon_file,
                 language,
                 template_style,
+                package_name,
```
kjrokos, line 380: I don't think we want to keep adding language-specific parameters here. Perhaps we could have a class representing each language that can prompt for and store any language-specific values it needs?

kjrokos: Also, I don't see `package_name` being set if the language is not Java -- which I think will error.

vmware_aria_operations_integration_sdk/mp_init.py
```diff
@@ -355,7 +377,7 @@ def main() -> None:
                 vendor,
                 eula_file,
                 icon_file,
-                language,
+                adapter_config,
                 template_style,
```
kjrokos, line 381: Since the available template are going to be (potentially) language-dependent, we should consider moving that (and the prompt for them) into the AdapterConfig class in a future PR.

vmware_aria_operations_integration_sdk/adapter_config.py
```diff
@@ -0,0 +1,26 @@
...
+
+class AdapterConfig:
+    def __init__(self, language: str, questions: List[Question] = []) -> None:
+        self.language = language
+        self.values: Dict[str, Any] = dict()
+        self.questions = questions
+
+    def prompt_config_values(self) -> None:
+        for question in self.questions:
+            self.values[question.key] = question.ask()
```
kjrokos, line 26: This is nice. I think as we refactor the init code, anywhere where we're branching on AdapterConfig::language is a good candidate to move the method into this class 👍

kjrokos (approved):

PR #307, opened 2023-10-04
Update documentation

quirogas:
Remove Python project specific references to make getting started more generic.

### DIFF
--- a/docs/get_started.md
+++ b/docs/get_started.md
@@ -37,13 +37,12 @@ In addition, at least one Cloud Proxy (also version 8.10 or later) must be set u
 * Git 2.35.0 or later. Updating to the latest stable version is recommended.
   For instructions in installing git, go to [Git's installation documentation](https://git-scm.com/downloads),
   and follow the instructions provided for your operating system.
-* Java JDK 17 (Only for Java Adapters).
-  As the Adapter operates within a container, developers don't need Java in their local environment.
-  Nevertheless, incorporating Java into your local development environment can facilitate the development process.
-  We suggest using JDK 17 when creating Java adapters, as combining JDK 21 or Other JDK versions may result in features 
-  incompatible with the container environment in which the Adapter operates and may have some compatibility issues.
-  To access the most recent JDK 17 release,
-  please visit [Java Downloads](https://www.oracle.com/java/technologies/downloads/#java17).
+* Azul Zulu JDK 17 (Only for Java Adapters).
+  Java Adapters compile and run within a container,
+  so a JDK is not strictly necessary, but it is generally useful for development.
+  The container is built with Zulu JDK 17, so ideally this version should be used for maximum compatibility.
+  To access the most recent Zulu JDK 17 release,
+  visit [azul.com](https://www.azul.com/downloads/?version=java-17-lts&package=jdk#zulu).
 
 ### Installation
 


docs/get_started.md
```diff
@@ -116,11 +126,10 @@ the creation of a new management pack project.
...
 
 ### Project Templates
-Both templates create a file system that has the basic project structure required to develop and build a Management Pack.
-Each file and directory is discussed in depth in the [mp-init](references/mp-init.md) documentation. `app/adapter.py` is the adapter's
-entry point and the best starting point.
+All the project templates create a file system that has the basic project structure required to develop and build a Management Pack.
+Each file and directory is discussed in depth in the [mp-init](references/mp-init.md) documentation.
 
-The methods inside `adapter.py` are required, and should be modified to generate a custom
+The methods the adapter source code are required, and should be modified to generate a custom
```
kjrokos, line 132: This sentence needs to be reworked a bit

docs/get_started.md
```diff
@@ -116,11 +126,10 @@ the creation of a new management pack project.
      see the [MP Initialization Tool Documentation](references/mp-init.md).
```
kjrokos, line 126: We should have a language-specific section here for the Java Package question

docs/get_started.md
kjrokos: Maybe we even push the descriptions of the methods to a different (language-specific) file, as the signatures are slightly different.

kjrokos (commented):
We probably want to have an optional requirement for the Java 17 JDK - it's not required even when writing a Java Adapter, but it's beneficial.

docs/get_started.md
```diff
@@ -113,66 +123,108 @@ the creation of a new management pack project.
...
 
 ### Project Templates
-Both templates create a file system that has the basic project structure required to develop and build a Management Pack.
-Each file and directory is discussed in depth in the [mp-init](references/mp-init.md) documentation. `app/adapter.py` is the adapter's
-entry point and the best starting point.
+All the project templates create a file system that has the basic project structure required to develop and build a Management Pack.
+Each file and directory is discussed in depth in the [mp-init](references/mp-init.md) documentation.
+To learn more about the specifics of each language: 
+    - Python 
+    - Java
```
kjrokos, line 135: This seems unfinished, and possibly redundant. Are the files going to be discussed in `mp-init`, or in a language-specific place?

docs/get_started.md
```diff
@@ -113,66 +123,108 @@ the creation of a new management pack project.
...
+    - `getAdapterDefinition()`:
+      Optional method that defines the Adapter Instance configuration. The Adapter Instance
+      configuration is the set of parameters and credentials used to connect to the target and
+      configure the adapter. It also defines the object types and attribute types present in a
+      collection. Setting these helps VMware Aria Operations to validate, process, and display
+      the data correctly. If this method is omitted, a `describe.xml` file should be manually
+      created inside the `conf` directory with the same data. Generally, this is only necessary
+      when using advanced features of the `describe.xml` file that are not present in this method.
+ 
 For further guidance on using the sample adapter, consult the `Guides` section.
```
kjrokos, line 221: Let's move this section out of get-started

docs/get_started.md
```diff
@@ -98,81 +100,64 @@ the creation of a new management pack project.
...
+
+    For the purposes of this *Get Started* guide, select **Sample Adapter**.
 
 ???+ info
 
      For complete documentation of the `mp-init` tool, including an overview of its output,
-     see the [MP Initialization Tool Documentation](references/mp-init.md).
+     see the [MP Initialization Tool Documentation](references/mp-init.md). 
+     For Java projects documentation consult [Java project](references/python_project/index.md) section   
+     For Python projects documentation consult [Python project](references/java_project/index.md) section
```
kjrokos, line 140: I'm thinking these links may be better at the conclusion of the walkthrough. Also the wording is a bit awkward.

docs/get_started.md
```diff
@@ -37,6 +37,8 @@ In addition, at least one Cloud Proxy (also version 8.10 or later) must be set u
 * Git 2.35.0 or later. Updating to the latest stable version is recommended.
   For instructions in installing git, go to [Git's installation documentation](https://git-scm.com/downloads),
   and follow the instructions provided for your operating system.
+* Java 17 or later (Only for Java Adapters).
+  For the latest release of Java visit [Java Downloads](https://www.oracle.com/java/technologies/downloads/)
```
kjrokos, line 41: I think we want to be more specific here:
* Even for Java Adapters, it's not required, just recommended (for ease of development)
* We want the JDK (rather than just the JRE)
* Versions above 17 may have features that won't work in our server. We could say something like "The latest release of Java 17"

docs/get_started.md
```diff
@@ -98,81 +100,64 @@ the creation of a new management pack project.
...
+     see the [MP Initialization Tool Documentation](references/mp-init.md). 
+     For Java projects documentation consult [Java project](references/python_project/index.md) section   
+     For Python projects documentation consult [Python project](references/java_project/index.md) section   
 
 ### Project Templates
-Both templates create a file system that has the basic project structure required to develop and build a Management Pack.
-Each file and directory is discussed in depth in the [mp-init](references/mp-init.md) documentation. `app/adapter.py` is the adapter's
-entry point and the best starting point.
+All the project templates create a file system that has the basic project structure required to develop and build a Management Pack.
+Each file and directory is discussed in depth in the [mp-init](references/mp-init.md) documentation.
```
kjrokos, line 144: We just linked to `mp-init` documentation a few lines ago. Let's remove one of them. Actually, this whole `Project Templates` section is feeling a bit out of place. We definitely want the information here, but maybe not in the 'get started' page.

We talked a bit about having a "Next Steps" section at the conclusion of the walkthrough. I think we can put a list of documents there that will give some guidance about how to start creating their own adapter without interrupting the flow of doing this walkthrough.

docs/get_started.md
kjrokos: Also, let's link to the Zulu distribution we use in the container. In addition to being the same version, I believe the license is more permissive than the Oracle version.

kjrokos (approved):

PR #312, opened 2023-10-11
Documentation/mp init java ammendment

### LINKED ISSUE #262: Expand the mp-init page to include Java portion

quirogas:
- closes #262

### DIFF
--- a/docs/references/mp-init.md
+++ b/docs/references/mp-init.md
@@ -78,7 +78,7 @@ An icon can be added later by setting the 'pak_icon' key in 'manifest.txt' to th
 [//]: # ( ❯ Python)
 [//]: # (```)
 
-### Language 
+#### Language 
 
 Adapters can be written in the following languages:
 


docs/get_started.md
```diff
@@ -110,32 +110,51 @@ the creation of a new management pack project.
...
 
-    `Python`
+    - `Python`
  
-    `Java` 
+    - `Java` 
+ 
+9.  === "Python"
+ 
+        a. `Select a template for your project`
```
kjrokos, line 119: Can we keep the numbering here consistent? It looks strange the way it's currently laid out:
image

I would expect the selector itself to be un-indented, and then inside an ordered list with numbers, starting at 9.

kjrokos: It seems like this might not be possible out of the box, but I found a plugin that purports to allow it: https://github.com/paulmelis/pm_attr_list

Which I found from this issue:
https://github.com/mkdocs/mkdocs/discussions/2838

docs/get_started.md
```diff
@@ -110,32 +110,51 @@ the creation of a new management pack project.
...
+
+           `New Adapter`: Creates methods, minimal comments, and code necessary to implement test
+           connection, collection, adapter definition, and endpoints logic.
+
+!!! warning "" 
 
     For the purposes of this *Get Started* guide, select **Sample Adapter**.
 
+---
+
```
kjrokos, line 157: The way this is formatted doesn't read as part of the step it is for. I think it would be better to duplicate and have it as part of each language's respective step. I'm also not sure that putting it in a warning box is appropriate, even without the 'warning' text/icon

docs/references/.pages
```diff
@@ -1,6 +1,8 @@
 nav:
   - Python Adapter Library: python-lib
   - Java Adapter Library: references/java-lib/index.html
+  - python_project
+  - java_project
```
kjrokos, line 5: image

I think we should either move the adapter library documentation in here, or just link directly to the page without the folder

docs/references/java_project/index.md
```diff
@@ -1,17 +1,96 @@
+# Overview 
+
+
+### Templates
+
+All the templates generate the base project structure along with some additional files and directories specific to
```
kjrokos, line 6: Might be good to link back to the base project structure description as well - similarly for the python overview

docs/references/java_project/index.md
```diff
@@ -1,17 +1,96 @@
+# Overview
```
kjrokos, line 1: Needs a more descriptive title

docs/references/java_project/index.md
```diff
@@ -1,17 +1,96 @@
...
+```
+
+### src (directory)
+
+This directory contains the package structure specified during `mp-init`. Inside the package, every template creates an
+Adapter.java file that serves as the main class for the project. 
+
+### build.gradle.kts (file)
+
+This files contains all the project dependencies along with some building specifications used by the Dockerfile.
```
kjrokos, line 57: We might want just a bit more detail here -- specifically, we should call out the tasks that we use for getting dependencies from the build stage to the final image (Maybe want to add a comment in the gradle file itself as well?)

docs/references/java_project/index.md
```diff
@@ -1,17 +1,96 @@
...
+
+### build.gradle.kts (file)
+
+This files contains all the project dependencies along with some building specifications used by the Dockerfile. 
+
+### commands.cfg (file)
+This file contains a list of the commands the HTTP server can run, along with the path to the executable related to the
+command. By default, all commands are run by executing the `adapter.py` file along with a parameter that defines a command.
+For example, when the HTTP server receives a request to run a test connection, it reads the commands.cfg key for `test`
+and runs the process defined by the key value, `/usr/local/bin/python app/adapter.py test`.
```
kjrokos, line 63: We should adapt this to the specifics of Java, or just put it in the base documentation

docs/references/java_project/index.md
```diff
@@ -1,17 +1,96 @@
...
   A typical test connection will generally consist of:
 
-    1. Read identifier values from adapterInstance that are required to connect to the target(s)
-    2. Connect to the target(s), and retrieve some sample data
-    3. If any of the above failed, return an error, otherwise pass.
-    4. Disconnect cleanly from the target (ensure this happens even if an error occurs)
+  1. Read identifier values from adapterInstance that are required to connect to the target(s)
+  2. Connect to the target(s), and retrieve some sample data
+  3. If any of the above failed, return an error, otherwise pass.
+  4. Disconnect cleanly from the target (ensure this happens even if an error occurs)
```
kjrokos, line 94: For some reason these nested lists aren't rendering correctly:
image

docs/references/mp-init.md
```diff
@@ -11,7 +11,7 @@ an initial project structure and create classifiers that other tools and VMware
 
 
 ## Prerequisites
-* The [VMware Aria Operations Integration SDK](../get_started.md#installation) is installed, with the virtual environment active.
+* The [VMware Aria Operations Integration SDK](../get_started.md#installation) is installed, with `pipx` (recommended).
```
kjrokos, line 14:
```suggestion
* The VMware Aria Operations Integration SDK is [installed](../get_started.md#installation).
```

docs/references/mp-init.md
```diff
@@ -78,20 +78,25 @@ An icon can be added later by setting the 'pak_icon' key in 'manifest.txt' to th
 [//]: # ( ❯ Python)
 [//]: # (```)
 
-##  Output
+### Language 
+
+Every project has the same inner server infrastructure; however, adapters can be written in other languages.
+Currently, we support the following languages:
```
kjrokos, line 84:
```suggestion
Adapters can be written in the following languages:
```

docs/references/mp-init.md
```diff
@@ -78,20 +78,25 @@ An icon can be added later by setting the 'pak_icon' key in 'manifest.txt' to th
...
+
+- Java
+- Python
+
+##  Output (Base Project)
```
kjrokos, line 89:
```suggestion
##  Base Project Structure
```

docs/references/python_project/index.md
```diff
@@ -1,17 +1,99 @@
+# Overview
```
kjrokos, line 1: Most of the comments on the Java project file apply here too

docs/references/java_project/index.md
kjrokos, line 1: The structure of this file (and the python one) can be improved a bit, I think. It currently looks like this:
image

I think the outline should look something like this:
```
# Java Project 
## Project Structure
  (keep content that was under 'templates' here - but I don't think the files/directories 
  need to be headers -- maybe a definition list (https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown/#definition-lists)?)
## Templates
### Sample Adapter
### New Adapter
## Adapter methods
```

docs/references/java_project/index.md
```diff
@@ -25,11 +104,11 @@ adapter. Each method fulfills a request from the VMware Aria Operations collecto
 
 - `collect(adapterInstance)`:
```
kjrokos, line 105: For the Java version, I think we should use the full signatures for the methods, e.g., `public CollectResult collect(AdapterInstance adapterInstance)`

kjrokos (commented):
Overall, the content is looking good - just need to shore up some formatting quirks with mkdocs, and a few organization issues.

docs/references/java_project/index.md
kjrokos: These might work well as definition lists as well -- see comment here: https://github.com/vmware/vmware-aria-operations-integration-sdk/pull/312#discussion_r1355355598

docs/get_started.md
```diff
@@ -110,32 +110,51 @@ the creation of a new management pack project.
...
 
-    `Sample Adapter`: Creates a template adapter that collects several objects and metrics from the container the adapter is running.
-    The template adapter has comments throughout its code that explain what the code does and how to customize it for your adapter.
+           The package name will be used to set up the package used by the adapter and the directory structure of the project.
+           
+        ???+ example 
+     
+            ``` java
+            package com.mycompany;
+            ```
```
kjrokos, line 140: Having an example here is nice, but I think it may be better to not have the java code and instead have the text we input for the question. If someone literally entered `package com.mycompany;` for this question, it wouldn't work.

kjrokos: Actually, surprisingly `mp-init` doesn't complain, but you end up with this:

image

docs/get_started.md
kjrokos: This is much better 👍

kjrokos (approved):

PR #314, opened 2023-10-16
Documentation/faq java ammendment

### LINKED ISSUE #264: Add Troubleshooting and FAQs for Java-related issues

quirogas:
- Reformat files
- Revise FAQ
- Add Java examples where applicable

Closes #264

### DIFF
--- a/docs/troubleshooting_and_faq/logs.md
+++ b/docs/troubleshooting_and_faq/logs.md
@@ -74,27 +74,37 @@ Each log file is prepended with the type of log file followed by a number that r
 
 === "Java Adapter Library"
 
-    Adapter logging is setup via the [Get Logger](../references/java_project/java_lib/-integration-s-d-k-adapter-library/com.vmware.aria.operations/get-logger.html) method
-    available through the `AdapterLogger` class of the Java Integrations SDK. The **Get Logger** method returns a Log4j Logger:
+    Adapter logging is setup via the [Setup Logging](../references/java_project/java_lib/com/vmware/aria/operations/AdapterLogger.html#setupLogging(java.lang.String,java.lang.Integer,java.lang.Long)) method
+    available through the `AdapterLogger` class of the Java Integrations SDK. The logger only need to be set up once.
+    To retrieve the logger, you can call the `Get Logger` static method:
 
     ```java linenums="1" hl_lines="3"
     import org.apache.logging.log4j.Logger;
     import com.vmware.aria.operations.AdapterLogger;
 
-    public void myMethod() {
-      Logger logger = AdapterLogger.getLogger();
-      logger.info("info log");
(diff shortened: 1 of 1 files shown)


docs/troubleshooting_and_faq/collections.md
```diff
@@ -7,33 +7,56 @@ Internal server errors can happen for various reasons; however, the most common
...
-
-  ```
+`mp-test` runs a series of validations test after a collection; if the collection has no results, then each validation
+step will report the result as missing.
+When a collection result is missing, it usually means an error occurred during a collection,
+but the Adapter handled the error.
+When the Adapter handles an error, the response contains an error message; The console displays the error message.
+For example:
+
+=== "Python Integration SDK"
```
kjrokos, line 20: Python Adapter Library/Java Adapter Library

docs/troubleshooting_and_faq/build.md
```diff
@@ -1,31 +1,21 @@
...
-  mp-build
-  Building adapter [Finished]
-  Unable to build pak file
-  ERROR: Unable to build Docker file at /Users/user/code/management-packs/test:
-   The command '/bin/sh -c pip3 install -r adapter_requirements.txt --upgrade' returned a non-zero code: 1
-  ```
-The solution for case 1 to fix the typo/command by editing the Dockerfile. For case 2, however, the solution might not be evident at first sight. Since the error
-comes from building the image itself, we can run `docker build .` in the project's root directory and look at the stack trace for clues.
+In most cases, this error indicates issues with building the container image. The most probable cause is:
+
```
kjrokos, line 6: We used to have a list of causes, you removed the list but didn't add the remaining cause to this sentence.

docs/troubleshooting_and_faq/logs.md
```diff
@@ -40,55 +50,100 @@ log file followed by a number that represents rollover.
...
+      logger.info("info log")
+      logger.warning("warning log")
+      logger.error("error log")
+      logger.debug("debug log")
+       ...
+    ```
 
+=== "Java Adapter Library"
 
+    Adapter logging is setup bia the [Get Logger](../references/java_project/java_lib/-integration-s-d-k-adapter-library/com.vmware.aria.operations/get-logger.html) method
```
kjrokos, line 77:
```suggestion
    Adapter logging is setup via the [Get Logger](../references/java_project/java_lib/-integration-s-d-k-adapter-library/com.vmware.aria.operations/get-logger.html) method
```

kjrokos: It's actually setup by the `setupLogging(...)` method (same as Python Adapter). Generally the user won't have to call that, because it's part of the `main` method we provide, but we should still be accurate.

kjrokos (approved):

PR #358, opened 2024-03-06
Fixes not properly parsing certificates in Java Adapter library

ahmed-sadak:
Closes #365

### DIFF
--- a/lib/java/AdapterLibrary/build.gradle.kts
+++ b/lib/java/AdapterLibrary/build.gradle.kts
@@ -12,7 +12,7 @@ plugins {
 }
 
 group = "com.vmware.aria.operations"
-version = "1.0.2"
+version = "1.0.3"
 
 java {
     toolchain {
--- a/lib/java/AdapterLibrary/src/test/java/com/vmware/aria/operations/AdapterInstanceTest.java
+++ b/lib/java/AdapterLibrary/src/test/java/com/vmware/aria/operations/AdapterInstanceTest.java
@@ -1,13 +1,106 @@
 package com.vmware.aria.operations;
 
-import kotlinx.serialization.json.*;
-import org.junit.jupiter.api.*;
+import kotlinx.serialization.json.JsonArray;
+import kotlinx.serialization.json.JsonElement;
+import kotlinx.serialization.json.JsonElementKt;
+import kotlinx.serialization.json.JsonNull;
+import kotlinx.serialization.json.JsonObject;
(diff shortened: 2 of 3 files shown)


lib/java/AdapterLibrary/build.gradle.kts
```diff
@@ -98,6 +98,6 @@ publishing {
     }
 }
 
-signing {
-    sign(publishing.publications["mavenJava"])
-}
+//signing {
+//    sign(publishing.publications["mavenJava"])
+//}
```
kjrokos, line 103: We'll need this to publish to Maven. It shouldn't prevent you from building locally - but either way, shouldn't be committed.

lib/java/AdapterLibrary/src/test/java/com/vmware/aria/operations/AdapterInstanceTest.java
```diff
@@ -116,6 +101,30 @@ private JsonPrimitive nil() {
...
+    }
+
+    private JsonPrimitive n(Number i) {
+        return JsonElementKt.JsonPrimitive(i);
+    }
+
+    private JsonPrimitive nil() {
+        return JsonElementKt.JsonPrimitive((Boolean) null);
+    }
+
```
kjrokos, line 127: I'm curious why you moved these

kjrokos (changes requested):
The lint check that's failing isn't related to your code. It's probably better fixed in a separate PR, so you can ignore for now.

lib/java/AdapterLibrary/build.gradle.kts
ahmed-sadak: my bad, forgot to revert this after testing locally

lib/java/AdapterLibrary/src/test/java/com/vmware/aria/operations/AdapterInstanceTest.java
ahmed-sadak: these were moved by my default formatter

kjrokos: Ah, that makes sense. Is that just the default IntelliJ settings?

ahmed-sadak: Yes, if there are any specific ones for this project I can import them, reformat and resubmit

kjrokos (approved):

PR #368, opened 2024-05-30
Fix documentation around adding traversals

kjrokos:
Issue:

The Integration SDK moved traversal spec definitions from `describe.xml` to `content/traversalSpecs/traversal.xml` but didn't update the corresponding documentation.

Resolution:

Updates the documentation to point to the correct files.


MarkArbogast (approved):
Approved pending CI issue resolution.

PR #374, opened 2024-06-11
Always stop timers when exiting the 'Timer.time' block

### LINKED ISSUE #361: com.vmware.aria.operations.time should stop the timer even if fn throws
### Describe the bug

If the `fn` passed to `com.vmware.aria.operations.time` throws an exception, the timer will not be stopped.

### Reproduction steps

This will not stop the *Login* timer and it probably should. If the exception is caught higher up the stack, the timer will continue running which is not desirable.
```kotlin
time("Login") { throw RuntimeException("Invalid credentials") }
```

### Expected behavior

If `fn` throws, the timer should be stopped.

### Additional context

_No response_

kjrokos:
Previously, throwing an exception inside the Timer.time() function block would cause the running timer to continue outside of the block.

This change does two things:
* When the block exits for any reason, the timer will stop
* Adds a `@Throws` annotation to the function block so Java code calling this method is also allowed to throw checked exceptions. Previously Java code could only throw unchecked exceptions, while Kotlin code was able to throw any exception. This brings the functionality closer to parity.

Resolves #361


TrueFiskibility:
Looks good. Note that Java code was always able to throw unchecked exceptions, but it wasn't able to throw checked exceptions. I suppose the choice to allow checked exceptions is a matter of design; I have no opinion on it.

MarkArbogast (approved):

PR #398, opened 2025-01-09
Explicitly exit at end of execution

### LINKED ISSUE #359: Java SDK testConnection additional latency of ~1min added
### Describe the bug

A JDK-based management pack with Integration SDK: when mp-test with 'test connection' is executed and the actual time taken is around 30-60s above the time taken inside both Java main method executions for 'get endpoints' and 'test connection' combined together

### Reproduction steps

1. Write a JDK-based mgmt pack which makes an additional call to an external system
2. Try `mp-test` and connect to that external system
3. With my example 'get endpoints' takes about 5 secs inside the main method in Java
4. 'test connection' takes about 16 secs inside the main method in Java
5. The above times out
6. Test with 'mp-test connect -t 90' which takes around 75 seconds -> looks like some minute or so is added by the framework and Python server
...

### Expected behavior

Expected behaviour is that there is not much additional delay apart from the actual execution of the Java main methods

### Additional context

_No response_

kjrokos:
Resolves #359 

Notes from investigation:
* OkHttp has the hanging issue, but the exact same code (as close as Ktor allows, which is quite close) using the Cio engine does not.
* In both cases, the server gets the result back in approximately the same amount of time. The hanging occurs after all other execution is finished.
* In the OkHttp case, the culprit appears to be an "OkHttp Dispatcher" thread that is still running. I've found a good number of issues related to this, but all from around ~2020 and claim be resolved in the versions we are using (e.g., https://youtrack.jetbrains.com/issue/KTOR-396)
* I tried manually closing the client, engine, dispatcher, executor, connection pool, etc - with no success. As near as I can tell, closing just the client should be sufficient, but clearly isn't in our case
* If I run the exact same code outside of a container, I don't have this problem. I don't know what is different about the container environment. It's also possible (maybe?) that there is a thread sticking around longer but it's not noticeable because the system itself isn't waiting on it to close. But I'm not sure how to test that, and it seems unlikely.
* If we explicitly call 'System.exit(0)' at the end of collection, everything works as expected. I don't love it, but this is likely the best (and most general) solution. The user shouldn't be doing anything at this point - any cleanup should have already happened, all results have already been sent back to the server, and the user would expect that the program exits here immediately anyway.
* I don't know if the issue can present itself in Python in the same way, but I have added exits at the same place in the Python templates for consistency

### DIFF
--- a/vmware_aria_operations_integration_sdk/adapter_configurations/adapter_templates/python/new_adapter/app/adapter.py.template
+++ b/vmware_aria_operations_integration_sdk/adapter_configurations/adapter_templates/python/new_adapter/app/adapter.py.template
@@ -143,7 +143,7 @@ def main(argv: List[str]) -> None:
         # `inputfile` and `outputfile` are always automatically appended to the
         # argument list by the server
         logger.error("Arguments must be <method> <inputfile> <ouputfile>")
-        exit(1)
+        sys.exit(1)
 
     method = argv[0]
     try:
@@ -161,13 +161,13 @@ def main(argv: List[str]) -> None:
                 logger.info(
                     "get_adapter_definition method did not return an AdapterDefinition"
                 )
-                exit(1)
+                sys.exit(1)
         else:
             logger.error(f"Command {method} not found")
-            exit(1)
+            sys.exit(1)
     finally:
         logger.info(Timer.graph())
-        exit(0)
+        sys.exit(0)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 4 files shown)


vmware_aria_operations_integration_sdk/adapter_configurations/adapter_templates/python/new_adapter/app/adapter.py.template
```diff
@@ -167,6 +167,7 @@ def main(argv: List[str]) -> None:
             exit(1)
     finally:
         logger.info(Timer.graph())
+        exit(0)
```
MarkArbogast, line 170: https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python

Should we use sys.exit() instead of exit() throughout?

MarkArbogast (changes requested):

MarkArbogast (approved):

